PR #17, opened 2023-03-23
RCO base app

ravermame:
1. Contains base libs
2. Login
3. User crud
4. Reports list


src/modules/LoginPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const [password, setPassword] = useState('');
+    const login = useLogin();
+    const notify = useNotify();
+    const [create] = useCreate();
+
+    const { data, total, isLoading } = useGetList('users');
+    const handleSubmit = (e: any) => {
+        e.preventDefault();
+        // will call authProvider.login({ email, password })
+        const userData = data?.filter((user: any) => user.name === username && user.password === password)
```
IanMayo, line 17: the above line should be `find`, not `filter`. Right?

1. We want the loop to drop out as soon as a match is found
2. Our handling is simpler if the value is a valid user, or undefined.

IanMayo: Aah, second note on this code. I'm pretty sure the username/password checking should be done in the `login` method of the `authProvider` - rather than having the logic here in the page code.

ravermame: If I want to move it to `authProvider` then I cannot use the `useGetList` hook(RA)  for retrieving data from `DataProvider`. 

In my understanding, this works if:
1. we have an actual API
2. or we bypass the default functions of `ra-data-local-storage` and fetch data from `localStorage` 

Thoughts @IanMayo  ?

IanMayo: Would ra-data-localforage solve the issue? https://github.com/marmelab/react-admin/tree/master/packages/ra-data-localforage

IanMayo:
Closing.

PR #18, opened 2023-03-23
Review of Salony code

IanMayo:


src/utils/index.ts
```diff
@@ -0,0 +1,32 @@
...
+	LIST_USERS = 'list_users',
+	LIST_AUDIT = 'list_audit',
+}
+
+export const trackEvent = (
+	type: AuditType,
+	activity_detail?: string,
+	defaultUser?: User
+) => {
+	const userData = localStorage.getItem('token');
```
IanMayo, line 19: @salony75way - ideally this code would not be aware that we are using `localStorage`.  It would be more versatile if it was dealing with the `dataProvider`.

So, when we move away from `localStorage` to a real REST backend, the code would still work.

IanMayo: Or, if this code is getting the token from localStorage - I would rather it was wrapped in another utility method. `trackEvent` should not have to know how we store user tokens.

src/providers/dataProvider/index.ts
```diff
@@ -0,0 +1,45 @@
+import localStorageProvider from 'ra-data-local-storage';
+import { withLifecycleCallbacks } from 'react-admin';
+import constants from '../../constants';
+import { AuditType, trackEvent } from '../../utils';
+import users from './users';
+
+export const dataProvider = localStorageProvider({
```
IanMayo, line 7: @salony75way - I think it's more complex to configure initially, but I would like us to try `ra-data-localforage`, since it seems a more modern, versatile way of using local-storage.  
https://github.com/marmelab/react-admin/tree/master/packages/ra-data-localforage

src/providers/dataProvider/index.ts
```diff
@@ -0,0 +1,45 @@
...
+	localStorageKey: constants.LOCAL_STORAGE_DB_KEY,
+	defaultData: {
+		users,
+	},
+});
+
+export const dataProviderWithLifeCycle = withLifecycleCallbacks(dataProvider, [
+	{
```
IanMayo, line 15: @salony75way - we don't need to audit `get` events, only data changes.

src/utils/index.ts
salony75way: Done.

src/providers/authProvider/index.ts
```diff
@@ -0,0 +1,56 @@
+import { AuthProvider } from 'react-admin';
+import constants from '../../constants';
+import { AuditType, trackEvent } from '../../utils';
+import userData from '../dataProvider/users';
+
+const authProvider: AuthProvider = {
+	login: (params) => {
+		const { username, password } = params;
+		const { users } = JSON.parse(
```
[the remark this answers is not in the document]
salony75way: Done.

src/providers/dataProvider/index.ts
salony75way: Done.

src/providers/dataProvider/index.ts
salony75way: Done.

IanMayo (commented):
Approved.

Issue #5, opened 2023-03-20
Provide some reference tables


IanMayo:
There are a number of drop-downs that are used in the interface.

We can usefully provide  support for these.  They will typically take a counter as a unique id, plus a single text value.

Here are the tables:
* Protective Marking
* Protective Marking Authority
* Platform Originator Class
* Vault
* Department
* Organisation

We should provide default data rows for these tables. There should be 5 rows in each, taking values "Protective-Marking-1", "Protective-Marking-2", and so on.

Only users with "adminRights" should be able to view or edit these tables.

We will use the content of these tables to populate drop-downs later in the project.

We can _probably_ avoid having these in the left-hand menu.  We can _probably_ have a "Reference Data" entry on the left-hand menu that opens a "Reference Data" page. This page shows a list of the reference tables, with an overview of what each represents.  The user can click on any to view their "List" view.

PR #34, opened 2023-03-27
Provide some reference tables

salony75way:
Provide some reference tables
#5


IanMayo (approved):
Merging. Will come back to TypeScript errors

PR #44, opened 2023-03-30
Secure storage

TahaKhanAbdalli:
User id secured in local storage upon user login and salt for that encryption is also stored in local storage


src/utils/ecnryption.ts
```diff
@@ -0,0 +1,12 @@
+import CryptoJS from 'crypto-js'
+const key = import.meta.env.VITE_SALT
```
IanMayo, line 2: could this have a fallback string, please - so it works in environments where we haven't (yet) created the environment variable.

IanMayo:
It _looks like_ we are still storing the password in the cookie:
image

IanMayo (approved):

PR #45, opened 2023-03-30
Password protect

TahaKhanAbdalli:
Password of users protected in localforage db and also password removed from being stored in localstorage as token


IanMayo:
Ok, I _almost_ understand what is happening.

But, `getToken()` should return the full user details object, so we can access the `adminRights` field. We can then delete the `adminRights` property when we push the object to the cookie.

TahaKhanAbdalli:
@IanMayo for the getToken comment. If we delete the adminRights field before pushing to the object to the cookie, then it will disturb our routing mechanism because the auth provider calls getToken and from the retun values it checks if the user has admin rights or not. if we are gonna delete that then we will not be able to determine if the user is admin or not.

TahaKhanAbdalli:
Yes we can @IanMayo. I'll fix the warning coming in the console.

IanMayo:
> @IanMayo for the getToken comment. If we delete the adminRights field before pushing the object to the cookie, then it will disturb our routing mechanism because the auth provider calls getToken and from the retun values it checks if the user has admin rights or not. if we are gonna delete that then we will not be able to determine if the user is admin or not.

But, if we store the whole of the `User` object in the encrypted object, we'll be able to determine the `adminUser` status, won't we?  So `getToken` would just decrypt the whole `User` object and return it.

TahaKhanAbdalli:
Yes got it we can encrypt whole of the object. Doing it will let you know once done.

TahaKhanAbdalli:
I am currently utilizing the bcryptjs library to encrypt and compare passwords for added security measures. This library makes use of the crypto module which is primarily designed for server-side operations, which is why the warning is being triggered. I have not come across any other library that performs this specific function. One possible solution is to apply the same encryption process as we use for local storage and subsequently decrypt and compare the user's password against the stored password in the database.

IanMayo:
> I am currently utilizing the bcryptjs library to encrypt and compare passwords for added security measures. This library makes use of the crypto module which is primarily designed for server-side operations, which is why the warning is being triggered. I have not come across any other library that performs this specific function. One possible solution is to apply the same encryption process as we use for local storage and subsequently decrypt and compare the user's password against the stored password in the database.

Thanks Taha.  Would you mind doing a little research on this?  We don't need `bomb-proof` security, just enough to dissuade a bored staff member playing with the app.  Could you see if there is a client-side lib (or even pair of methods) that we could use with the `salt` value to encrypt/decrypt the user login details?

src/providers/authProvider/index.ts
```diff
@@ -4,24 +4,22 @@ import { AuditType, trackEvent } from '../../utils/audit'
...
+  if (encryptedUser !== null && salt !== null) {
+    const decryptedData = decryptData(`${encryptedUser}`)
+    return decryptedData.substring(0, decryptedData.length - 32)
   } else return null
 }
 
-const setToken = (token: string): void => {
+const setToken = (token: string, salt: string): void => {
   localStorage.setItem(constants.TOKEN_KEY, token)
+  localStorage.setItem(constants.SALT, salt)
```
IanMayo, line 16: But @TahaKhanAbdalli - doesn't thevalue of `salt` come from the environment variable?  Why are we storing it in the cookie?

src/providers/authProvider/index.ts
```diff
@@ -2,25 +2,23 @@ import { type AuthProvider, type DataProvider } from 'react-admin'
...
-  if (JsonUser !== null) {
-    const user = JSON.parse(JsonUser) as User & { salt: string }
-    const decryptedData = decryptData(`${user.id}${user.salt}`)
-    user.id = decryptedData.substring(
-      0,
-      decryptedData.length - user.salt.length
-    )
-    return JSON.stringify(user)
-  } else return null
+export const getUser = (): string | undefined => {
```
IanMayo, line 5: could this return an instance of `User` please?  Then the rest of the app doesn't need to know how we store the user data.

src/providers/authProvider/index.ts
TahaKhanAbdalli: It is general practice to store salt along with the encrypted data. What we are getting from env is the key.

IanMayo: Sure, ok @TahaKhanAbdalli

src/providers/authProvider/index.ts
TahaKhanAbdalli: It was already there (returning string) but I can return user.

IanMayo: `getUser = (): User | undefined`

IanMayo:
For the `Codacy` errors @TahaKhanAbdalli  - you need to explicitly declare the members of the template as strings.  (or, you may be able to provide another solution)
image

TahaKhanAbdalli:
ok I understand @IanMayo

IanMayo:
Nice work, thanks @TahaKhanAbdalli 

There just seems one bit remaining: to be able to view/edit the password in `UserEdit`

TahaKhanAbdalli:
> Nice work, thanks @TahaKhanAbdalli
> 
> There just seems one bit remaining: to be able to view/edit the password in `UserEdit`

Got it @IanMayo . Will work on it.

IanMayo:
@TahaKhanAbdalli  - when I add a user, the password doesn't get encrypted:
image

I don't think this is a symptom of us temporarily having an in-memory database - the password for new users still still need to be encrypted even when we move to formal database.  Aah, but maybe that will happen on the server.

Anyway - let's get this issue sorted, then we can move on.

TahaKhanAbdalli:
I have also encrypted password on user create.

PR #52, opened 2023-03-31
Support for adding items to batches

### LINKED ISSUE #9: Support for adding items to a batch
Tens or hundreds of items can be added to a batch.

The "flow" for adding an item to a batch, is to find the batch, and then add the item(s).

The most recent batches will be shown on the dashboard, but users can search for them too.

Once a batch has been found, then "Add item" will create a new item, pre-assigned to that batch.

The item will also contain these fields:
* media type: foreign key
* [ a set of fields specific to the media type ]
* start: date
* end: date
* vault location: text
* remarks
* protective marking

Since staff will be entering a long series of these, buttons should be provided to "Save and add another (blank)" and "Save and add another (copy this)".  These will create a blank new item (attributed to this batch), or a duplicate of the current one if it's a series of similar items.

Note: I don't have a set idea for how the different media types are handled. I guess we _could_ use the `TabbedForm" to have one set of fields per media type.

salony75way:
fixes #9


IanMayo:
Hello @salony75way  - I've just had a play with adding an item to a batch.  I see that the new item doesn't have the `batch_id` assigned.  Could it be that the URL is expressing `batch=0` instead of `batch_id=0`?

IanMayo:
Hello @salony75way - nice progress :-)

I've just added a new item, and it does now set the correct batch-id.  For the short-term I think we need to provide evidence that the item is being assigned to the correct batch. 

So, I think that on the ItemCreate form, we need to show a read-only field for Batch Reference Number (`V03/2023`).  I've updated the UX to show this:
https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g221ed9988a1_1_133

We won't need to show `batch reference` for `ItemList`, since the `item reference` actually includes the `batch reference`.

IanMayo:
I know it's a work in progress, but a quick bit of feedback.  The fields on the "media" tabs for "itemedit" should not be required.  They are mutually exclusive, so a user will never have to enter data for more than one media tab.

IanMayo (approved):

IanMayo:
Accepting, since `Codacy` not providing details of failure.

PR #55, opened 2023-04-04
batch id generated based on the batches of the selected year

TahaKhanAbdalli:


src/providers/dataProvider/index.ts
```diff
@@ -62,6 +62,23 @@ export const getDataProvider = async (): Promise<DataProvider<string>> => {
   )
   const providerWithCustomMethods = { ...provider }
   const audit = trackEvent(providerWithCustomMethods)
+
```
IanMayo, line 65: 1. Could you provide a return type for the method please?
2. Why not filter the records in the initial `getList`?

IanMayo (approved):
Great, thanks/

PR #58, opened 2023-04-05
Changed all reference data to resources

TahaKhanAbdalli:


IanMayo (changes requested):
Thanks @TahaKhanAbdalli , now I understand the strategy.

image


But I don't want to _pollute_ the left-hand side bar with reference data.  I just want it to contain a couple of specific resources that users will access 99% of the time.

Is it possible to return to just displaying a few resources in the side-menu?
image

src/resources/department/DepartmentList.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { List, TextField, TopToolbar, Datagrid } from 'react-admin'
+
+const ListActions = () => <TopToolbar></TopToolbar>
+
+export default function DepartmentList(): React.ReactElement {
+  return (
+    <List actions={<ListActions />}>
```
IanMayo, line 8: Thanks @TahaKhanAbdalli  - for these reference data lists, we don't need `show` behaviour, since the user can see the name if the entry in the `list` view. We just need `edit` behaviour.  

Hmm, switching to an `edit` form just to edit one string still seems cumbersome.  Is it possible to edit the value in a small popup modal?  If the user is editing multiple reference items, they will still have the list in the background, and will still "have their eyes on" the rows they are going to edit.

IanMayo:
Closing this PR.  We're going to adopt a "lighter" strategy for how we handle these entities.

PR #63, opened 2023-04-05
reference data edit and create implemented

TahaKhanAbdalli:


IanMayo:
Can we go straight to `Edit` please?   The `Show` form is a step in the process that doesn't bring any value.

TahaKhanAbdalli:
Should I completely remove the show page?

IanMayo:
> Should I completely remove the show page?

Yes please.  The `list` shows the same data as `show`.  We just need `edit` and `create` (which can prob use the same form :-) ).

IanMayo:
@TahaKhanAbdalli -  there's something that `Codacy` doesn't like, too:
image

IanMayo (approved):
Great, thanks.

PR #66, opened 2023-04-06
created mock data for project, batch and platform

### LINKED ISSUE #13: Create body of mock data
We're starting with a minimal dataset.  This is right, since it will require minimal refactoring as the data schema matures.

But, once the data schema is mature, we should generate a higher volume of fictional data.

Modify `DataProvider` to create: 
- 10 platforms
- 10 projects (spanning years 2020 to 2023, taking random platforms)
- 10 batches
- 20 items in each batch

Note: it should be relatively easy to "tune" the numbers above, so we can generate a high volume of data, for testing - and to find bottlenecks in the future.

We do have TypeScript definitions for these types, so when we create the mock data we should use strong typing.

We typically have 5 or more entries in each of our reference tables.  It's just the row-id that is stored in the above element types. So, for reference fields, we really just have to generate a random integer between 0 and 5.

For free text fields, just use the name of the field ('Title') plus an integer counter.  So, `Project Names` will be `Name:1`, `Name:2`, etc.

In order to support time-related searches, the time fields in the data need to have some consistency:
- end dates are all after start dates
- item dates fit within the parent project dates

Project and Batches will have a time length of 4 to 10 weeks.  Item dates will have a time length of 1 to 10 days

TahaKhanAbdalli:
Fixes #13

### DIFF
--- a/src/utils/generateData.ts
+++ b/src/utils/generateData.ts
@@ -3,7 +3,9 @@ import { DateTime } from 'luxon'
 const MediaType = ['DVD', 'Tape', 'Paper']
 
 function generateRandomNumber(min: number, max: number): number {
-  const randomNumber = Math.random()
+  const array = new Uint32Array(1)
+  const generatedRandomNumber = window.crypto.getRandomValues(array)
+  const randomNumber = generatedRandomNumber[0] / (Math.pow(2, 32) - 1)
   const scaledRandomNumber = randomNumber * (max - min) + min
   const randomInteger = Math.floor(scaledRandomNumber)
   return randomInteger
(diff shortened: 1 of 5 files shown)


src/utils/generateData.ts
```diff
@@ -0,0 +1,153 @@
...
+      new Date(project.start_date),
+      new Date(project.end_date)
+    )
+    const startDate = generateRandomDateInRange(
+      new Date(project.start_date),
+      new Date(endDate)
+    )
+    const id: number = generateRandomNumber(1, 100000)
+    const obj: Item = {
+      id,
```
IanMayo, line 130: Hello @TahaKhanAbdalli - for all the objects we create here, we should leave the `id` blank, so that the DataProvider generates it.
We generate random numbers for two of the types, but they may allow duplicate `id` values.
Aah, maybe you are providing the `id` field because TypeScript wants it.

One solution would be to declare a `Partial<Item>`.  But that would allow multiple fields to be missing - and we want TypeScript to force us to provide values for all required fields.  So, instead we could declare a type that [omits](https://stackoverflow.com/a/63991735/92441) the id, and create an instance of that type.

(note: this is right at the edge of my knowledge of TypeScript 🤣 )

TahaKhanAbdalli:
I can do that but while creating items I have to have the id to generate `item_number` property. Should I assign random string generated by `item-number-${i}`  to `item_number`?

IanMayo:
> I can do that but while creating items I have to have the id to generate `item_number` property. Should I assign random string generated by `item-number-${i}` to `item_number`?

Yes, a random number would be fine :-)

IanMayo (approved):

PR #68, opened 2023-04-07
61 created at field

TahaKhanAbdalli:


src/providers/dataProvider/index.ts
```diff
@@ -100,11 +101,21 @@ export const getDataProvider = async (): Promise<DataProvider<string>> => {
...
+        const { data } = record
+        const { id } = data
         await audit(
           AuditType.CREATE_PROJECT,
           `Project created (${String(record.data.id)})`
         )
+        await dataProvider.update<Project>('projects', {
+          id,
+          previousData: data,
+          data: { created_at: DateTime.now().toFormat('yyyy-MM-dd') }
```
IanMayo, line 117: @TahaKhanAbdalli - could you put the data creation into a reusable function?  Then, if we decide we want to introduce time too, we only have to do it in one location.

src/providers/dataProvider/index.ts
```diff
@@ -100,11 +101,21 @@ export const getDataProvider = async (): Promise<DataProvider<string>> => {
...
+        })
         return record
       },
       afterUpdate: async (record: UpdateResult<Project>) => {
```
IanMayo, line 121: @TahaKhanAbdalli - just to help my react-admin skills....  Is there a reason why you inject the `created_at` data in the `afterCreate`, rather than `beforeCreate`?

IanMayo:
Thanks @TahaKhanAbdalli  - there is a lot of complexity in the `created at` search.  Could we remove that from this PR please?  Let's just concentrate on inserting (and displaying) the created date.

IanMayo (approved):

PR #74, opened 2023-04-10
Store time for items

### LINKED ISSUE #67: Store time for items
Some items contain data that lasts several days. Other items have data just covers a few hours. For the latter, it would be useful to store the time component of the start/end too.

For our mock data, we should generate time values in multiples of 15 minutes (10:15, 22:45, 16:00, etc)

So, for display/edit of items, we should let the date AND time be viewed/edited.

I'm pretty sure this will involve the adoption of this component:
https://marmelab.com/react-admin/DateTimeInput.html

salony75way:
fixes #67

### DIFF
--- a/src/resources/items/ItemForm/CoreForm.tsx
+++ b/src/resources/items/ItemForm/CoreForm.tsx
@@ -1,6 +1,10 @@
-import { ReferenceInput, SelectInput, TextInput } from 'react-admin'
+import {
+  DateTimeInput,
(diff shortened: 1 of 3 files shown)


IanMayo:
Hello @salony75way - the minutes are being rounded to 15 minutes, but we still have random seconds.  You should set the seconds to zero, please.
image

IanMayo (approved):

PR #94, opened 2023-04-12
name filter on project list

### LINKED ISSUE #78: Search for ProjectList
We should introduce search/filter for Projects.

We need free-text-search on Project Name, drop-down list for year of Project Start Date

* Phase 1 is free-text-search
* Phase 2 is filter by year

TahaKhanAbdalli:
Fixes #78

### DIFF
--- a/src/resources/projects/ProjectList.tsx
+++ b/src/resources/projects/ProjectList.tsx
@@ -11,7 +11,7 @@ import {
   TopToolbar
 } from 'react-admin'
 
-const filters = [<SearchInput source='name' key='name' alwaysOn />]
+const filters = [<SearchInput source='q' key='q' alwaysOn />]
 
 export default function ProjectList(): React.ReactElement {
   const ListActions = () => (
--- a/src/resources/projects/ProjectList.tsx
+++ b/src/resources/projects/ProjectList.tsx
@@ -4,20 +4,25 @@ import {
   CreateButton,
   Datagrid,
   DateField,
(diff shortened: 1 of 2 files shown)


IanMayo:
Hmm, I'm having a play.  If I filter for `project` - I don't get any matches.

The `FakeRest` [lib](https://github.com/marmelab/FakeRest) that this data-provider sits on _does_ provide partial matches:
image

Could you investigate and see if we can fix this?  Maybe try logging the `DataProvider`?

src/resources/projects/ProjectList.tsx
```diff
@@ -11,7 +11,7 @@ import {
   TopToolbar
 } from 'react-admin'
 
-const filters = [<SearchInput source='name' key='name' alwaysOn />]
+const filters = [<SearchInput source='q' key='q' alwaysOn />]
```
IanMayo, line 14: Hello @TahaKhanAbdalli - would just this change have fixed the issue?

Do we need to modify the data provider?

TahaKhanAbdalli: Yes, It by default searches all the properties but in the issue filter was specified on the name field therefore I tried to add only on the name field

IanMayo: Aah yes.  _For-now_, I'm tempted to stick with the `q` wildcard search.

IanMayo (approved):

PR #97, opened 2023-04-12
generateBatchId fixed

### LINKED ISSUE #86: Fix Batch Generation issue
Our `batch numbers` include a counter for the relevant year.

We currently generate the counter by getting all batches for that year, then returning the length of the list.

But, if we delete a batch, we will re-generate the current last batch number.
- we have two batches `V01/2023` and `V02/2023`
- a user deletes `V01/2023`
- when we add a new batch, it will be called `V02/2023` - so we have to batches with that batch reference number

Instead, we should get all of the batches, extract the highest counter number, and return an incremented version of it.

TahaKhanAbdalli:
Fixes #86

### DIFF
--- a/src/providers/dataProvider/index.test.ts
+++ b/src/providers/dataProvider/index.test.ts
@@ -148,9 +148,10 @@ describe('generateBatchId for values greater than 9', () => {
           })}/${year}`
         )
       }
-      const ids = []
       // deleting first 10 entries
-      for (let i = 0; i < 10; i++) ids.push(i)
+      const ids = Array(10)
+        .fill(null)
+        .map((_, i) => i)
       await provider.deleteMany(constants.R_BATCHES, { ids })
 
       // inserting new 10 entries
--- a/src/providers/dataProvider/index.test.ts
+++ b/src/providers/dataProvider/index.test.ts
@@ -28,8 +28,10 @@ const mockProvider = {
     batches[resource].data = []
   },
 
-  async delete(resource: string, toDelete: { id: number }) {
-    batches[resource].data.splice(0, toDelete.id)
+  async deleteMany(resource: string, toDelete: { ids: number[] }) {
+    batches[resource].data = batches[resource].data.filter(
+      (b) => !toDelete.ids.includes(b.id)
+    )
   }
 }
 
@@ -90,19 +92,19 @@ describe('generateBatchId', () => {
   })
 
   describe('when there is one batch in the specified year', () => {
-    it('should return 01', async () => {
+    it('should return 02', async () => {
       await generateBatch(id, provider, year)
       const result = await generateBatchId(provider, year)
-      expect(result).toBe('01')
+      expect(result).toBe('02')
     })
   })
 
   describe('when there are multiple batches in the specified year', () => {
-    it('should return 02', async () => {
-      await generateBatch(id, provider, year)
+    it('should return 03', async () => {
+      await generateBatch(id++, provider, year)
       await generateBatch(id, provider, year)
       const result = await generateBatchId(provider, year)
-      expect(result).toBe('02')
+      expect(result).toBe('03')
     })
   })
 
@@ -146,9 +148,10 @@ describe('generateBatchId for values greater than 9', () => {
           })}/${year}`
         )
       }
-
+      const ids = []
       // deleting first 10 entries
-      await provider.delete(constants.R_BATCHES, { id: 10 })
+      for (let i = 0; i < 10; i++) ids.push(i)
+      await provider.deleteMany(constants.R_BATCHES, { ids })
     })
   })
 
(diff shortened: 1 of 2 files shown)


IanMayo:
It looks like the implementation works, but we need to allow for the `V` number to go above 9.  I don't think this code would handle `V12`

src/providers/dataProvider/index.test.ts
```diff
@@ -72,7 +72,7 @@ describe('generateBatchId', () => {
     it('should return 00', async () => {
       provider.clear(constants.R_BATCHES)
       const result = await generateBatchId(provider, year)
-      expect(result).toBe('00')
+      expect(result).toBe(`0${id}`)
```
IanMayo, line 75: I can see you have changed these since the `id` changes.  But, I do want the results to be `00`, `01` and `02` - since I/we can see that they are the results that should be generated.

I _think_ you just have to initialise the `id` at the start of the test.

Hmm, since we want to reset the `id` _AND_ clear the batches, they could be in a jest `beforeAll` method.

TahaKhanAbdalli:
> It looks like the implementation works, but we need to allow for the `V` number to go above 9. I don't think this code would handle `V12`
> 
> Actually, could you extend the suite that creates batch numbers up to 20, and creates the correct batch id? Aah, then do a test that `delete`s the first 10 batches for that year, and re-generate the id? The generated id should still be `20`.

This code is handling values greater than 9.

IanMayo:
> This code is handling values greater than 9.

Great - can you please add a test that demonstrates this?

src/providers/dataProvider/index.test.ts
```diff
@@ -104,3 +115,69 @@ describe('generateBatchId', () => {
...
+          `V${i.toLocaleString('en-US', {
+            minimumIntegerDigits: 2,
+            useGrouping: false
+          })}/${year}`
+        )
+      }
+
+      // deleting first 10 entries
+      await provider.delete(constants.R_BATCHES, { id: 10 })
+    })
```
IanMayo, line 152: The above code looks like it is deleting batch with `id` of 10, rather than the first 10 entries.

The important part of the test comes next though. When we generate a new batch ID, we should see that it provides the value of 20 (max id) rather than 10 (current length)

src/providers/dataProvider/index.ts
```diff
@@ -36,10 +48,27 @@ export const generateBatchId = async (
     pagination: { page: 1, perPage: 1000 },
     filter: { yearOfReceipt: year }
   })
-  return batches.data.length.toLocaleString('en-US', {
-    minimumIntegerDigits: 2,
-    useGrouping: false
-  })
+
```
IanMayo, line 51: hmm, for the next line - I understand why we have to handle a zero length array (since `reduce` will fail).  But, we don't have to handle `1` here, do we?  I _guess_ we can just let the `reduce` method run on the single item.

The pros and cons are probably 50/50 on it, but I _think_ I favour just using this `special handling` for an empty array.  Aah, plus...  if we're just handing a zero length array, we can return `00`, without considering the array length.

src/providers/dataProvider/index.test.ts
TahaKhanAbdalli: The length is being passed as id due to typescript as delete in provider has this type of arguments. But it is working ok in the mock provider delete function. The next test is generating new 10 batches and then verifying if the new batches have proper ids i.e `V20` on-wards.

IanMayo: Aah, I see the issue.  In that case, you should use `deleteMany`, and provide a list of ids to delete.

src/providers/dataProvider/index.test.ts
```diff
@@ -28,8 +28,10 @@ const mockProvider = {
     batches[resource].data = []
   },
 
-  async delete(resource: string, toDelete: { id: number }) {
-    batches[resource].data.splice(0, toDelete.id)
+  async deleteMany(resource: string, toDelete: { ids: number[] }) {
```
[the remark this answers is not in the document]
TahaKhanAbdalli: It is but for the sake of testing I created a mock provider in the test file which just pushes data to an array and mocks the provided `getList`, `clear` and `deleteMany` functions. so no actual data provider is being used. Therefore, `deleteMany` is just removing elements from the array.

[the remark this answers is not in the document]
IanMayo: Aah, I should have been paying more attention ;-)

src/providers/dataProvider/index.test.ts
```diff
@@ -104,3 +117,70 @@ describe('generateBatchId', () => {
...
+          })}/${year}`
+        )
+      }
+      const ids = []
+      // deleting first 10 entries
+      for (let i = 0; i < 10; i++) ids.push(i)
+      await provider.deleteMany(constants.R_BATCHES, { ids })
+    })
+  })
+
```
IanMayo, line 157: aah, now I see what is happening. This test relies on the outcome of the previous test (that the  10 items were deleted).

Tests should not rely on each other.

I _guess_ you could just _move_ the contents of the second test into the first test.

IanMayo (approved):

PR #103, opened 2023-04-14
99 default data

TahaKhanAbdalli:


IanMayo:
Nice refactoring.  But, when the app starts on a vanilla device, do we still generate the default data?

IanMayo (approved):

PR #113, opened 2023-04-14
Improved item show

### LINKED ISSUE #101: Improved ItemShow
We should improve how we support ItemShow

We should have a read-only version of the ItemCreate form - with a main details page, plus supporting tabs.

salony75way:
fixes #101


IanMayo (changes requested):
Thanks @salony75way 

1. Can we have the `item` reference number in the view header bar, like you did for `batches` please?
image

2. Using a disabled version of the controls is a good `low-maintenance` solution, but the data isn't as legible as I'd like.  Is it an option to use some different styling in `show` mode, so they controls don't have a gray background.
image

IanMayo:
Thanks for the update :-)

* [ ] We need the text to be a darker shade, it's still not properly legible (see below)
* [ ] ItemShow seems to be lacking an `Edit` button
* [ ] There is some issue in opening an `Item` from it's parent `Batch`.  The app is returning `Element does not exist`.

image

IanMayo:
Thanks for the update @salony75way 
* [ ] The ItemShow form still isn't properly legible, the text is light-gray, it needs to be a dark/black shade (see below)
* [ ] The form menu should show the `item_number`. (V02/2023/04), not the name of the batch it came in (`batch-1`). I wanted you to copy the presentation from the `reference` in the `BatchShow` form, not the value itself.
* [ ] I'd like the `ItemEdit` to also show the `item_number` in the header bar, just like in updated `ItemShow`.

image

IanMayo (changes requested):
Thanks @salony75way - but, it looks like for items we are showing the batch number (V02/2023) not the item number (V02/2023/05).

salony75way:
> Thanks @salony75way :-)
> 
> Does the `BatchNumber` component need all that complexity? It _seems_ to be doing the same job as `ItemNumber`.
> 
> Hmm, I wonder if it's possible to just add another `prop` to `ItemNumber` for `source`, then _maybe_ it could work for Batches and Items?
> 
> Aah, could you also address the `Codacy` warning too, please?

ATM we can use BatchNumber component for reference field, batch from query params and in batch show. So let me know if we don't need  these functionality

salony75way (commented):
Revert Update TopToolbarField.tsx

IanMayo (approved):

PR #116, opened 2023-04-17
security related field added in audit table

### LINKED ISSUE #81: Introduce "Security-related" audit events
We should introduce a `securityRelated` flag to audit events.

This will be `true` if the event relates to the `protectiveMarking` or `numberOfPages` for an `Item`.

This will have to be part of the `ItemEdit` process - checking if the `protectiveMarking` has been changed, and setting the flag accordingly.

We don't have `numberOfPages` yet - we'll handle the flag for those once they are created.  For this phase we'll just handle changes to the protective marking.

TahaKhanAbdalli:
Fixes #81

### DIFF
--- a/src/providers/dataProvider/index.ts
+++ b/src/providers/dataProvider/index.ts
@@ -127,22 +127,12 @@ export const getDataProvider = async (): Promise<DataProvider<string>> => {
     {
       resource: constants.R_BATCHES,
       beforeUpdate: async (record: UpdateParams<Batch>) => {
-        if (
+        await audit(
+          AuditType.EDIT_BATCH,
+          `Batch updated (${String(record.data.id)})`,
           record.previousData.maximumProtectiveMarking !==
-          record.data.maximumProtectiveMarking
-        ) {
-          await audit(
-            AuditType.EDIT_BATCH,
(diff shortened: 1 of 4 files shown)


src/providers/dataProvider/index.ts
```diff
@@ -125,6 +126,25 @@ export const getDataProvider = async (): Promise<DataProvider<string>> => {
...
+        if (
+          record.previousData.maximumProtectiveMarking !==
+          record.data.maximumProtectiveMarking
+        ) {
+          await audit(
+            AuditType.EDIT_BATCH,
+            `Batch updated (${String(record.data.id)})`,
+            record.previousData.maximumProtectiveMarking !==
+              record.data.maximumProtectiveMarking
+          )
```
IanMayo, line 139: hmm, the above `if` has already been tested at aline 131.  It can just be `true` here, right?

IanMayo: Change implemented in this commit: https://github.com/DeepBlueCLtd/RCO/pull/116/commits/99e5dc4479b89bf9761bb5e90dec2cd7ccdb5c2e

PR #153, opened 2023-04-21
Introduce createdBy field for Items, Batches, and Projects

### LINKED ISSUE #125: Introduce `createdBy` field for Items, Batches, and Projects
Frequently one user will be responsible for adding a whole set of data.  While processing it, they will consider it as "my data" - and frequently refer to it.

So, we should introduce `createdBy` for `Projects`, `Batches` and `Items`.   

* [ ] In the `Show` pages for these we will show the `user-name` for the `createdBy` field.
* [ ] For the `List` pages we will introduce two filters.  One will be `Created by` - which will show a drop-down list of users. The other will be a [Quick Filter](https://marmelab.com/react-admin/FilteringTutorial.html#quick-filters) with a default value of current user id, called `Created by me`.

salony75way:
fixed #125


src/components/CreatedByInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default function CreatedByInput(props: Props): React.ReactElement {
+  const { source = 'createdBy' } = props
+  const { data } = useGetIdentity()
+  const formContext = useFormContext()
+
+  useEffect(() => {
+    if (formContext !== null) {
+      const { setValue, getValues } = formContext
+
```
IanMayo, line 15: @salony75way - we should probably popup an error dialog if there isn't an identity, it's probably a but that would want to spot ourselves, or get users to tell us about.

IanMayo:
Hello @salony75way.  I don't feel strongly about this, but is there a reason why we don't handle `createdBy` in the `DataProvider` in `AfterCreate`?
By doing it there, we could _probably_ use one method in several places (for different resources).

salony75way:
> Hello @salony75way. I don't feel strongly about this, but is there a reason why we don't handle `createdBy` in the `DataProvider` in `AfterCreate`? By doing it there, we could _probably_ use one method in several places (for different resources).

Hi @IanMayo, `afterCreate` is a type of response intercepter so when we use this for data manipulation, we always need to call an additional api to update the data. And also for get identity, we always have to call an additional api so data will take time to save that's why i did't used this for createdBy

IanMayo:
> we always need to call an additional api to update the data

Aah yes, the row gets pushed to the database, then we have to make the change, and push it again.

So, what about in `beforeCreate`.  I _guess_ that if we inject `createdBy` there, the data hasn't yet gone to the database.

IanMayo:
>  And also for get identity, we always have to call an additional api so data will take time to save

Hmm, but identity is in the cookies, so this isn't a database round-trip.  It will effectively be instantaneous, won't it?

IanMayo:
@salony75way - could you fix the unit-tests too, please.  For the `generateData` stuff I guess we'll need to provide one of the existing user ids. 

Hmm, or we make `createdBy` optional?  No, let's not - so that TypeScript looks after the data for us.

IanMayo:
I'm happy to discuss this, it's just a thought...

image

We support `createdBy` using a custom filter component.  Couldn't we also allow this by providing a `SourceInput` that has a default value of `currentUser`?

I _guess_ it would look look like this

```
  <SourceInput
    key='createdByMe'
    source='createdBy'
    label='Created By Me'
    defaultValue={currentUserId}
    reference={constants.R_USERS}
  />,

  <SourceInput
    key='createdBy'
    source='createdBy'
    reference={constants.R_USERS}
  />,
```

salony75way:
> I'm happy to discuss this, it's just a thought...
> 
> image
> 
> We support `createdBy` using a custom filter component. Couldn't we also allow this by providing a `SourceInput` that has a default value of `currentUser`?
> 
> I _guess_ it would look look like this
> 
> ```
>   <SourceInput
>     key='createdByMe'
>     source='createdBy'
>     label='Created By Me'
>     defaultValue={currentUserId}
>     reference={constants.R_USERS}
>   />,
> 
>   <SourceInput
>     key='createdBy'
>     source='createdBy'
>     reference={constants.R_USERS}
>   />,
> ```

Yes we can do this but by default currentUser id will be undefined so we have to wait to load user before filter render, that's why i used custom filter here

IanMayo:
> Yes we can do this but by default currentUser id will be undefined so we have to wait to load user before filter render, that's why i used custom filter here

We can load currentUser id in an instant :-)
```
  const { data, isLoading } = useGetIdentity()
```

salony75way:
> > Yes we can do this but by default currentUser id will be undefined so we have to wait to load user before filter render, that's why i used custom filter here
> 
> We can load currentUser id in an instant :-)
> 
> ```
>   const { data, isLoading } = useGetIdentity()
> ```

No, this will take time to load and if we use default value after render then value will not update in input. Also we have to write this logic three times as we are using created by me filter in items, batches and projects.

So let me know if we still need to use  SourceInput field here.

IanMayo:
Aah, hold on @salony75way - this looks like the best UI for a `Created by me` filter:
https://marmelab.com/react-admin/FilteringTutorial.html#quick-filters

IanMayo (approved):

PR #157, opened 2023-04-21
Show empty list and fixed pagination conflicts with recent batches

### LINKED ISSUE #156: Can't add first `Item` to a `Batch`
When the `ItemList` component is empty, we don't have an `Add Item` toolbar button. We don't have a toolbar :-)

image

salony75way:
fixes #156 156

### DIFF
--- a/src/components/Recent.tsx
+++ b/src/components/Recent.tsx
@@ -96,6 +96,7 @@ export default function Recent<T>(props: Props<T>): React.ReactElement {
     <RecentCard label={label} resource={resource}>
       <ResourceContext.Provider value={resource}>
         <List
+          storeKey={`recent-${resource}` }
           hasCreate={false}
           actions={false}
           perPage={itemsCount}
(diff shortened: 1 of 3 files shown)


src/components/Recent.tsx
```diff
@@ -96,6 +96,7 @@ export default function Recent<T>(props: Props<T>): React.ReactElement {
     <RecentCard label={label} resource={resource}>
       <ResourceContext.Provider value={resource}>
         <List
+          storeKey={`recent-${resource}` }
```
IanMayo, line 99: @salony75way - can you clarify what the pagination problem was please?

IanMayo (approved):

src/components/Recent.tsx
salony75way: List params corresponding to resources are stored to local storage so when we open bathes list after welcome page then the recent batches pagination was applied to batch list and also batches are sorted by descending order.
This was vice versa so just changed the storeKey for recent batches.

PR #184, opened 2023-04-25
data provider logging added in user menu

### LINKED ISSUE #148: Add UI toggle for `logging enabled`
In development we may choose to switch on `loggingEnabled` for a build.

But, it may prove useful in production, too - certainly in the early stages.

Could we add a toggle in the user menu for `Logging enabled`. We'd store this in a react-admin pref, and use the value in the `DataProvider`

TahaKhanAbdalli:
Fixes #148

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,14 +37,35 @@ function App(): React.ReactElement {
   const [dataProvider, setDataProvider] = useState<DataProvider | undefined>(
     undefined
   )
-
+  const [loggingPref, setLoggingPref] = useState<null | string>(null)
   const handleGetProvider = (): any => {
-    if (dataProvider !== undefined) return
-    getDataProvider().then(setDataProvider).catch(console.log)
+    if (loggingPref !== null)
+      getDataProvider(loggingPref === 'true')
+        .then(setDataProvider)
+        .catch(console.log)
   }
 
-  useEffect(handleGetProvider, [dataProvider])
+  useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -37,14 +37,35 @@ function App(): React.ReactElement {
...
+        .catch(console.log)
   }
 
-  useEffect(handleGetProvider, [dataProvider])
+  useEffect(() => {
+    const storedValue = localStorage.getItem(constants.LOGGING_ENABLED)
+    if (storedValue !== null) {
+      setLoggingPref(storedValue)
+    } else {
+      setLoggingPref('true')
```
IanMayo, line 53: I _think_ the logging should be off by default, please.

src/App.tsx
```diff
@@ -37,14 +37,35 @@ function App(): React.ReactElement {
   const [dataProvider, setDataProvider] = useState<DataProvider | undefined>(
     undefined
   )
-
+  const [loggingPref, setLoggingPref] = useState<null | string>(null)
```
IanMayo, line 40: hello @TahaKhanAbdalli - while the value is stored as string in localStorage, I'd like to hide that from the rest of the app, please. So, I'd like the state to be boolean (default value false), and when we retrieve it from `localStorage` immediately convert to a boolean.

The big picture here is that the fact it's stored as a string in the browser is a detail that the rest of the app doesn't need to know about :-)

PR #186, opened 2023-04-25
asset report for items

### LINKED ISSUE #162: Asset Report for any group of items
We had a `muster report` for batch. It's not needed there.

But, we do need a similar `Asset Report` for any selection of `items` from `ItemList`.

This report could be used in a number of ways - but the general use case is when someone needs a list of items but they do not have access to the RCO system itself. So, the list would either be printed off or pdf'd and emailed to the recipient.

The report would be formatted as:
- title: `RCO Asset Report`
- header: list filters that were used to collate the listing, plus the current date/time and user 
- detail: list of items, including:
    - item number
    - media type
    - original reference
    - remarks

salony75way:
fixes #162

### DIFF
--- a/src/resources/items/ItemsReport.tsx
+++ b/src/resources/items/ItemsReport.tsx
@@ -1,12 +1,40 @@
-import { Typography } from '@mui/material'
-import React from 'react'
-import { Datagrid, List, TextField, type ListProps } from 'react-admin'
+import { Box, Button, Card, CardContent, Typography } from '@mui/material'
+import React, { useState } from 'react'
+import {
+  Datagrid,
+  List,
+  Show,
+  SimpleShowLayout,
+  TextField,
+  useListContext,
+  type ListProps
+} from 'react-admin'
 import SourceField from '../../components/SourceField'
 import * as constants from '../../constants'
+import { DateTime } from 'luxon'
+import { Download } from '@mui/icons-material'
+import Printable from '../../components/Printable'
 
-export default function ItemsReport(
-  props: Omit<ListProps, 'children'>
-): React.ReactElement {
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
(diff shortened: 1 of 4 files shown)


IanMayo:
Hello @salony75way - two comments based on the UI:
- `ItemList` is paged. But the report should show all matching rows (not just the first 10) 
- We have the `ItemList` in `BatchShow`.  When I ask for the AssetReport it shows the contents from the previous `ItemList` filter, not the set of items in this batch.

src/resources/items/ItemsReport.tsx
```diff
@@ -1,12 +1,40 @@
...
+import { DateTime } from 'luxon'
+import { Download } from '@mui/icons-material'
+import Printable from '../../components/Printable'
 
-export default function ItemsReport(
-  props: Omit<ListProps, 'children'>
-): React.ReactElement {
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
+
```
IanMayo, line 20: the previous two declarations seem generic. Should they go in a central place, to encourage their re-use?

salony75way: Ok i will move to types.d.ts file

IanMayo: Actually, can't we avoid using it by specifying an newer version of TypeScript?

IanMayo:
Nearly there @salony75way :-)

I have a batch open.  I'm looking at the assets.  I filter by protective marking - I see just the items in that batch at that protective marking. When I click on `Asset Report` I see all items from all batches with that protective marking, from all batches.

So in the `AssetReport` for `BatchShow` we need to add the `id` for the `Batch` as an implied filter.

IanMayo (approved):

PR #193, opened 2023-04-26
149 is active flag platform

### LINKED ISSUE #149: Use `isActive` flag for `Platforms`
We have added the `isActive` flag for platforms.

Here is how we should use it:
- in `BatchCreate` filter platforms drop-down for active platforms
- in `BatchList` provided separate filters for `Platform (all)` and `Platform (active)`.
- in `BatchEdit` provide full platforms drop-down (since it may refer to old/inactive platform)
- in `PlatformsList` on reference page, include filter for `isActive` - so maintainer can focus on active platforms, if necessary

Note: the steps above are probably similar to those needed for #140

TahaKhanAbdalli:
Fixes #149

### DIFF
--- a/src/resources/batches/BatchList.tsx
+++ b/src/resources/batches/BatchList.tsx
@@ -44,17 +44,16 @@ interface PlatformFilterType {
   label: string
   reference: string
   source: string
-  filter?: string
 }
 
 const PlatformFilter = (props: PlatformFilterType) => {
   const { data: batches } = useGetList('batches')
   const platformIds = batches?.map((batch) => batch.platform) ?? []
-  const { label, reference, filter = 'active' } = props
+  const { label, reference } = props
   const { setFilters, displayedFilters } = useListContext()
   const { data } = useGetMany(reference, { ids: platformIds })
   useEffect(() => {
-    if ((data != null) && filter === 'active') {
+    if (data != null) {
       const filteredData = data.filter((d) => d.active === true)
       setFilters(
         {
--- a/src/resources/batches/BatchList.tsx
+++ b/src/resources/batches/BatchList.tsx
@@ -92,7 +92,7 @@ const filters = [
     reference={constants.R_PLATFORMS}
     key='platform'
     sort={sort()}
-    source='platform'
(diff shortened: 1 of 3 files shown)


IanMayo:
This is working, but we could do with some help from you please @salony75way 

We have a [quick filter](https://github.com/DeepBlueCLtd/RCO/pull/193/files#diff-ddfe6681f4f6c7a67e86186ebf7b2bfaea2d788e8363d1d90fa3767dcd98f330R97) for active platforms.  But, we can't get a normal `Platform` filter to work.  If we add one filter, then both filters appear.

Steps to reproduce:
- open `BatchList`
- add `Active platforms` quick filter
- see that two platform filters are shown:
image

salony75way:
> This is working, but we could do with some help from you please @salony75way
> 
> We have a [quick filter](https://github.com/DeepBlueCLtd/RCO/pull/193/files#diff-ddfe6681f4f6c7a67e86186ebf7b2bfaea2d788e8363d1d90fa3767dcd98f330R97) for active platforms. But, we can't get a normal `Platform` filter to work. If we add one filter, then both filters appear.
> 
> Steps to reproduce:
> 
> * open `BatchList`
> * add `Active platforms` quick filter
> * see that two platform filters are shown:
>   image

This is because you are using same `source` key for both filter, You can get resolve this by using different `source` for one of then so here you can use `platform_eq` key.

src/resources/batches/BatchList.tsx
```diff
@@ -57,16 +89,16 @@ const filters = [
...
   <SourceInput
-    reference={constants.R_VAULT_LOCATION}
-    key='vaultLocation'
+    reference={constants.R_PLATFORMS}
+    key='platform'
     sort={sort()}
-    source='vaultLocation'
-  />,
-  <SourceInput
     source='platform'
```
salony75way, line 95: You can use `source="platform_eq"` here

IanMayo:
Thanks @salony75way - the fix worked :-)

IanMayo (approved):

PR #199, opened 2023-04-27
receiptNotes and remarks field made mandatory on batch create

### LINKED ISSUE #194: Batch Create - don't allow `Submit` with missing data
For BatchCreate it is possible to submit the form with missing fields.

This should not be possible.  We should prevent the new batch from being created if the form fails validation.

TahaKhanAbdalli:
Fixes #194


IanMayo:
Sorry - the issue was unclear.

I don't need `remarks` and `receipt notes` to be mandatory. I need the `Save` button to be disabled if mandatory fields are missing.

IanMayo:
Hmm, I've just re-tested, and the form is disabled. Closing issue and PR.

PR #201, opened 2023-04-27
Added url params to login

### LINKED ISSUE #174: URL param to login (temporarily)
It would be good to distribute a URL that allowed RCO to open, already logged in as a user.

This _could_ also be useful for E2E testing, if/when we have that.

TahaKhanAbdalli:
Fixes #174

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,8 @@ import {
   Resource,
   CustomRoutes,
   Loading,
-  type DataProvider
+  type DataProvider,
+  type AuthProvider
 } from 'react-admin'
 import { Route } from 'react-router-dom'
 import MyLayout from './components/Layout'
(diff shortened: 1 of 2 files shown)


IanMayo:
Thanks Taha, it works - but we get this warning.  Can you work out why it is displayed, and fix it please?
image

TahaKhanAbdalli:
This is by default being shown by react admin because react admin checks the authentication before mounting the component. I can add the delay until the toaster disappears but that will not be a good user experience.

IanMayo:
Hmm, is it an option to put the logic somewhere else?  Maybe in `app.tsx`?

TahaKhanAbdalli:
There is a trade off. I would have to show the notification of wrong password or any error manually (using material UI) rather than using `useNotify` of react admin.

IanMayo:
I'm afraid I won't understand the difference until I see a working implementation.

Could you have a go at putting it in `app.tsx` please?

TahaKhanAbdalli:
Pushed.

IanMayo (approved):

PR #202, opened 2023-04-27
E2e login test

IanMayo:
Introduce Playwright framework, add initial test


.gitignore
```diff
@@ -23,3 +23,10 @@ dist
 *.sln
 *.sw?
 .env
+node_modules/
+/test-results/
+/e2e/playwright-report/
```
IanMayo, line 28: @IanMayo  - if we ignore `/e2e` the child folders will be ignore too, won't they?

e2e/tests/login.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    await page.goto('/RCO')
+
+    const usernameInputField = page.locator('#username')
+    const passwordInputField = page.locator('#password')
+    const loginButton = page.locator('button[type="submit"]')
+    const invalidUserOrPasswordMessage = page.locator(
+      '//div[contains(@class, "MuiSnackbarContent-message")]'
+    )
+
+    await usernameInputField.type('ian')
```
IanMayo, line 16: Is this still a throwback to the POM strategy?

The above can be represented through this, can't it?
```
await page.locator('#username').type('ian')
```

I _think_ the above strategy would reduce about 20 lines from the file.  Yes, this is very much down to taste, and you have the experience, but I'd like to start with the smallest/simplest solution - and refactor to introduce complexity when it is required.

IanMayo:
One change required from the existing unit tests, please:
image

.gitignore
gitsachin: yes, if we just ignore `/e2e` the it'll ignore all child folders. In this case we are just ignoring `playwright-report` inside `e2e`

playwright.config.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig, devices } from '@playwright/test'
+export default defineConfig({
+  testDir: './e2e/tests',
+  fullyParallel: true,
+  forbidOnly: !(process.env.CI == null),
+  retries: process.env.CI != null ? 2 : 0,
+  workers: process.env.CI != null ? 1 : undefined,
+  reporter: 'html',
+  use: {
+    baseURL: 'https://deepbluecltd.github.io/',
```
IanMayo, line 10: @qajay97 - is the test running against the hosted instance at that URL?  We need the test to run against the code for the current branch.

Sorry, I have no idea how to do that, but I hope Google is your friend.

qajay97: Changed the baseUrl to include `/RCO` path

.gitignore
IanMayo: Duh, of course :-)

IanMayo:
Hello @qajay97 - we are still getting the GitHub Actions error described [above](https://github.com/DeepBlueCLtd/RCO/pull/202#issuecomment-1525431527)

It looks like you have to provide a `jest.config.ts` file, with instructions to ignore the `e2e` folder.

PR #207, opened 2023-04-27
Loan process 

### LINKED ISSUE #181: Loan process
# UI

UX presented here: https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g2375b13bec5_0_2

We would also add a `Loans` resource.  This would show in the left-hand menu.  The `List` would include these columns:
- created
- Holder
- Items Outstanding (calculated count)

The `Show` component is described in the above UI

# Logic:
I _guess_ we would implement this logic in the DataProvider.  We would add this new method:

```
loanItems => (items: Array<Item['id'], recipient: User['id'], by: User['id'], remarks: string): void
returnItems => (items: Array<Item['id'], by: User['id']): void
```

The method would perform the following steps:

## New loan
- create new `Loan` element
- create new `Loan Item` elements

## Return
- mark  `Loan Item` elements with `returnedDate`
- if all `Loan Item` elements for a `Loan` have been returned, mark `Loan` as returned

### Return process

- a user wishes to return a number of items
- the admin opens LoanList, and filters for loans for that user
- admin selects the relevant loan, viewed in LoanShow
- switch to `Loan Items` tab, which displays `LoanItemList`
- filters for `Items on loan`
- admin selects items being returned, then clicks on `Return`

# Data
```
Loan:
  - id: integer
  - createdAt: string
  - holder:  User['id']
  - loanedBy:  User['id']
  - remarks: string
 
LoanItem:
   - id: integer
   - createdAt: string
   - item: Item['id']
   - loan: Loan['id']
   - receivedBy: User['id']
   - returnedDate: string | null  // (null if still on loan)
   - remarks: string
```

salony75way:
fixes #181


IanMayo:
New requirement. please @salony75way - I think we need to know which user loaned the item to the borrower:
image

IanMayo:
Hello @salony75way - the linter is highlighting the fact that `Loan.by` is never read.  We should provide the `Loaned by` as a column and filter in `LoanList`.

IanMayo:
Thanks @salony75way - I've pulled the newest commit, but I don't see `Loan` button for `ItemList` or `BatchShow/ItemList`.
Am I doing something wrong?

salony75way:
> Thanks @salony75way - I've pulled the newest commit, but I don't see `Loan` button for `ItemList` or `BatchShow/ItemList`. Am I doing something wrong?
> 
> image

Let me check

IanMayo:
Aah, now that we have working UI, I understand more about what we need.

* [ ] For `LoanList` we should not have a `Create` button
* [ ] When making a loan, we don't need to enter `loaned-by`, it's the current user.  So we can just have a form with `Loan to` and `Remarks`.
* [ ] For the `ItemList` under `LoanShow` we should show the `ItemNumber` and `MediaType` (UX updated)
* [ ] For the `ItemList` we don't need to see `ReceivedBy` or `Loan`
* [ ] For the `ItemList` we should not have `Add Item` - this isn't a place where new items can be registered

IanMayo (approved):

PR #211, opened 2023-04-28
Added created at filter in item list and project list

### LINKED ISSUE #61: Include `created` field for high level entities
For some high level entities, we should include a `created` date.

This will let users quickly find items added in the past day/week/month.

This field should be added for `projects`, `batches` and `items`.  The field should be added to the TS interface definition for the tables.

Note: once we have `created` attribute for `items` we can add a filter for `Past week`, `Past month` and `Past year`.

I guess we _Could_ handle this through a field on the `create` form. But it would probably be better to handle it in the `create` handler in the `data-provider`.

**Update**: we have implemented the creation of `created_at` field, and now have to move onto using it in search/filters

TahaKhanAbdalli:
Fixes #61


IanMayo:
Could we have this filter for `Batches` too, please?

PR #214, opened 2023-04-28
Changed drop down filters to autocomplete

### LINKED ISSUE #145: Filter drop-down list that allows text to be typed in
In our filters, there are a number of places where we provide drop-down lists, typically to entities in other tables (foreign keys).

In the live data there are 10s and 10s of projects, platforms, users, and so on.  Scrolling down through the long list will be a chore.  It would be easier if the user could type in the label, and a wildcard-style matcher trim the list _live_ to matching entries.

To keep the processing down, we could wait until 3 chars have been pressed (or something like that).

It would be good to investigate if there is a way of providing the above "off-the-shelf".  If there isn't such a component, it would seem worth creating one.

TahaKhanAbdalli:
Fixes #145


IanMayo (changes requested):
One issue, please.  The text entry doesn't seem to work for `Projects` in `BatchForm`

TahaKhanAbdalli:
Can you please check again because it's working for me?

IanMayo:
> Can you please check again because it's working for me?

I get this strange behaviour: https://gyazo.com/09685244e6b52daa736151b27b3ec6c2

IanMayo (changes requested):
Conflict to resolve, please.

Let's get this merged - and come back to the `project` issue.

PR #221, opened 2023-05-01
default value set if there is only one item

### LINKED ISSUE #216: Provide default when only one item present
In the BatchCreate form, we will only have one item in several drop-down lists.

So, it's wasted effort for the user having to open the drop-down list.

If there is only one item present in the list for `Department` or `Organisation`, then we should set that as the default/current value.

TahaKhanAbdalli:
Fixes #216


IanMayo (changes requested):
Conflict to resolve here, please @TahaKhanAbdalli

IanMayo (changes requested):
Conflict to resolve too, please @TahaKhanAbdalli

IanMayo (approved):

PR #224, opened 2023-05-02
Decoupled activity type from displayed label

### LINKED ISSUE #128: Decouple `activity-type` value from displayed label
In the `Audit-Log` we currently display the JS label for the activity (`edit_item`).

But, IMHO we should de-couple the displayed label from the value.  This _could_ allow us to change the displayed value from the underlying data.

This does not need to be in a database table, having it hard-coded in source is ok.

So, we will introduce an `ActivityTypes` object that is a `ReferenceItem[]`.  We then populate it with the existing activity-type strings, but also add a human-formatted version of the text.  We will use this object for display/filter/search in `AuditList`.

Note: the previous paragraph is just a _guess_ at how to implement this feature.  I don't have any strong feelings about how we do it, but I don't want it in the DataProvider.

TahaKhanAbdalli:
Fixes #128

### DIFF
--- a/src/utils/activity-types.ts
+++ b/src/utils/activity-types.ts
@@ -19,19 +19,16 @@ export enum AuditType {
   EDIT_ITEM = 'edit_item'
 }
 
-const ActivityTypes = Object.values(AuditType).reduce<ActivityType[]>((acc, val) => {
+const ActivityTypes = Object.values(AuditType).map((val) => {
   const replacedVal: string = val.replace('_', ' ')
   const capitalizedFirstLetter: string = replacedVal.charAt(0).toUpperCase()
   const remainingLetters: string = replacedVal.substring(1).toLowerCase()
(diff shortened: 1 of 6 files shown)


src/utils/activity-types.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  CREATE_BATCH = 'create_batch',
+  DELETE_BATCH = 'delete_batch',
+  EDIT_BATCH = 'edit_batch',
+
+  CREATE_ITEM = 'create_item',
+  DELETE_ITEM = 'delete_item',
+  EDIT_ITEM = 'edit_item'
+}
+
```
IanMayo, line 21: Hey @TahaKhanAbdalli  - I'm certainly not the kind of guy who is afraid of `reduce` functions, but... ;-)

Can we handle this through a `map` instead?

IanMayo (approved):

PR #226, opened 2023-05-02
Refine e2e in Pull Requests

IanMayo:


IanMayo:
Thanks @qajay97 - but it "looks" like the e2e isn't being run.

image

qajay97:
@IanMayo In order to run e2e tests, we will need to add another step called `yarn e2e` after `yarn test`. That should execute our playwright test.

IanMayo:
Going to merge this, so we can continue to learn about `e2e`

PR #228, opened 2023-05-03
cache browser issue resolved

### LINKED ISSUE #191: How to refresh cached user layout on S/W update
I wish to change the default set of visible columns, and the default column sequence for a `ConfigurableDataList`.

But, for users that have already run the app, they will have a stored set of columns.

I personally am able to clear the cookies in my browser using Chrome-Dev-Tools. But, I can't expect users to do this.

How do we refresh these settings for users?

I _guess_ one option is to have a key for the component, and increment a counter on the key.  Then the browser will ignore the previous prefs for this table, and load new defaults

TahaKhanAbdalli:
Fixes #191


IanMayo (changes requested):
@TahaKhanAbdalli - do we need to add a `.env` file to the repo?

I _think_ it makes sense to have it checked in - then I/we can see when a PR is going to reset the IU or Data.

TahaKhanAbdalli:
So, should I push the .env also?

IanMayo:
> So, should I push the .env also?

Yes please. Unless you see a reason why we shouldn't?

.env
```diff
@@ -0,0 +1,2 @@
+VITE_DATA_VERSION="1.0.0"
```
IanMayo, line 1: I _don't think_ we have the concept of major/minor/patch in the UI/Data versioning. I presume just an integer counter would be sufficient.  At least for now..

IanMayo (approved):

PR #231, opened 2023-05-03
heroku issue fixed

### LINKED ISSUE #151: Issue with Heroku deployment
image

We get a Heroky deployment for each PR.  But, the deployment target doesn't work, it needs `/RCO/` to be added to the URL.

TahaKhanAbdalli:
Fixes #151

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: yarn preview --host 0.0.0.0 --base /RCO/ --port $PORT
+web: yarn preview --host 0.0.0.0 --base ./ --port $PORT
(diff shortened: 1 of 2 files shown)


IanMayo:
Note: I _think_ the solution will lie in this file: https://github.com/DeepBlueCLtd/RCO/blob/main/Procfile

IanMayo:
Aah, ok @TahaKhanAbdalli - so the fix was to remove the workaround I introduced while trying to fix the original problem.

Of course! 🤣

IanMayo (approved):

PR #235, opened 2023-05-04
Extended audit log to include user id, item and resource

### LINKED ISSUE #233: Audit Log extensions
The audit log can be used to support some business processes beyond security logging.

To do this we have to give it a richer data model.  Add these fields:
```
// the user making the change
user: User['id]
// the type of data being reported on (opt)
resource: string | null
// the id of the entity being reported on (opt)
id: number | null
```

Update the `trackEvent` method to support these new attributes. Note: I've made them `value | null` so that we are forced to consciously put relevant values (or null) in the `trackEvent` call - rather than to make them optional.

With the above data changes in place, we should extend `AuditList` filters (and columns) to include `User` and `Resource`.

TahaKhanAbdalli:
Fixes #233

### DIFF
--- a/src/providers/authProvider/index.ts
+++ b/src/providers/authProvider/index.ts
@@ -55,7 +55,12 @@ const authProvider = (dataProvider: DataProvider): AuthProvider => {
           const salt: string = generateSalt()
           const token = encryptData(`${JSON.stringify(clonedUser)}${salt}`)
           setToken(token, salt)
-          await audit(AuditType.LOGIN, 'Logged in')
+          await audit({
+            type: AuditType.LOGIN,
+            activityDetail: 'Logged in',
+            resource: constants.R_USERS,
+            item: null
+          })
           return await Promise.resolve(data)
         } else {
           throw new Error('Wrong password')
@@ -65,7 +70,12 @@ const authProvider = (dataProvider: DataProvider): AuthProvider => {
       }
     },
     logout: async (): Promise<void> => {
-      await audit(AuditType.LOGOUT, 'Logged out')
+      await audit({
+        type: AuditType.LOGOUT,
+        activityDetail: 'Logged out',
+        resource: constants.R_USERS,
+        item: null
+      })
       removeToken()
       await Promise.resolve()
     },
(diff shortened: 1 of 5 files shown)


src/providers/authProvider/index.ts
```diff
@@ -55,7 +55,12 @@ const authProvider = (dataProvider: DataProvider): AuthProvider => {
           const salt: string = generateSalt()
           const token = encryptData(`${JSON.stringify(clonedUser)}${salt}`)
           setToken(token, salt)
-          await audit(AuditType.LOGIN, 'Logged in')
+          await audit({
+            type: AuditType.LOGIN,
+            activityDetail: 'Logged in',
+            resource: constants.R_USERS,
```
IanMayo, line 61: Nice thinking, but let's provide `null` for `resource`, please - since we're not affecting the `users` table.

src/providers/authProvider/index.ts
```diff
@@ -65,7 +70,12 @@ const authProvider = (dataProvider: DataProvider): AuthProvider => {
       }
     },
     logout: async (): Promise<void> => {
-      await audit(AuditType.LOGOUT, 'Logged out')
+      await audit({
+        type: AuditType.LOGOUT,
+        activityDetail: 'Logged out',
+        resource: constants.R_USERS,
```
IanMayo, line 76: again, let's provide `null` for `resource`, please.

IanMayo:
Aah, I was unclear about the `item`.  It's actually the `id` - regardless of whether this is an item, a batch, or whatever.

Correction here  image

src/providers/dataProvider/index.ts
```diff
@@ -129,15 +129,30 @@ export const getDataProvider = async (
...
+        })
         return record
       },
       afterCreate: async (record: CreateResult<User>) => {
-        await audit(AuditType.CREATE_USER, `User created (${record.data.id})`)
+        await audit({
+          type: AuditType.CREATE_USER,
+          activityDetail: `User created (${record.data.id})`,
+          resource: constants.R_USERS,
+          item: null
```
IanMayo, line 145: aah, this should be `id`, not `item`, and it should take `record.data.id` - we can `eventually` list all audit events for this user record.

src/types.d.ts
```diff
@@ -19,6 +19,9 @@ interface User {
 
 interface Audit {
   readonly id: number
+  user: User['id']
+  resource: string | null
```
IanMayo, line 23: could you include the jsdoc comments from the original issue please?

IanMayo (changes requested):
Feedback provided, thanks.

IanMayo (approved):

PR #236, opened 2023-05-04
180 data provider tests - Phase 1

### LINKED ISSUE #180: Tests for data-provider
We're embedding some business logic in the data-provider.

We are validating part of this through the test for the `getBatchId()` logic.  

But, we should probably be testing the overall functionlity of the REST end-point to avoid regression issues.

We will _eventually_ switch to the SQLite backend, so it will probably be useful to be able to run the tests against the eventual backend.

The testing should probably include:
- putting database into predictable state
- look where there is logic in the Data Provider, and test the logic
- general CRUD operations on each resource

We should start off with a simple solution, to check things work. Find a DataProvider method that includes business logic, and check that logic in that method works as expected. 

Note: within this, we should also do the refactoring in #262

TahaKhanAbdalli:
Supports #180


IanMayo (changes requested):
There is a lot to review here @TahaKhanAbdalli 

Could we trim it back to testing just one resource (let's go for `batch`), and think through how we test it?

Once we have that sorted we can extend it to the other resources.

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      const updatedValue = generateDummyBatchForTesting('dummy-batch')
+      const result = await provider.update<Batch>(R_BATCHES, {
+        id,
+        previousData: batch,
+        data: {
+          name: 'dummy-batch',
+          createdAt: DateTime.now().toFormat('yyyy-MM-dd')
+        }
+      })
+      const { id: omittedId, ...expectedResult } = result.data
```
IanMayo, line 67: why is the id omitted?  
It should match the one passed in, shouldn't it?

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      expect(expectedResult).toMatchObject(updatedValue)
```
IanMayo, line 68: since we're updating a batch, shouldn't it be different in some way?

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    })
+
+    it('should delete the batch', async () => {
+      await expect(
+        provider.delete<Batch>(R_BATCHES, { id })
+      ).resolves.toBeTruthy()
```
IanMayo, line 74: shouldn't we verify that the item as been deleted?

src/providers/dataProvider/dataprovider.test.ts
TahaKhanAbdalli: The id is automatically generated by localforage when a new item is inserted. We can pass the id but I thought localforage should handle it iteself.

src/providers/dataProvider/dataprovider.test.ts
TahaKhanAbdalli: We update only the name of the batch and we also generate a new batch by passing the same name. So the updated value is name which we verify.

IanMayo: This is an `update` event - so the id should not change.

src/providers/dataProvider/dataprovider.test.ts
TahaKhanAbdalli: We are checking if the promise is resolved which is resolved if the item is deleted successfully .

src/providers/dataProvider/dataprovider.test.ts
IanMayo: Again - this is `update`, not `create`.  You have copy/pasted the test for `create` 🤣

src/providers/dataProvider/dataprovider.test.ts
IanMayo: Just to double-check, can we check that the `id` is no longer present?

The logic in the promise could change, and this test wouldn't tell is that it is no longer deleting.

src/providers/dataProvider/dataprovider.test.ts
TahaKhanAbdalli: For create the name is `Dummy-Batch` (capitalized). But when we update it is `dummy-batch` (small letters) and to verify the updated value we generate the batch by passing the name `const updatedValue = generateDummyBatchForTesting('dummy-batch')
` (small letters)

src/providers/dataProvider/dataprovider.test.ts
IanMayo: Aah yes - I misinterpreted the logic.  I understand how it is working now :-)

src/providers/dataProvider/dataprovider.test.ts
TahaKhanAbdalli: The last test tries to get the deleted batch which verifies if the batch is deleted.

IanMayo: Aah yes - can we move that `await` up into this test then please?

src/providers/dataProvider/dataprovider.test.ts
IanMayo: @TahaKhanAbdalli - I may be mistaken, but I still don't think we should omit the id.  If the id has changed, it is an indication that `update` (or `get`) is broken.

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      batch = result.data
+      expect(id).toBeDefined()
+      expect(batch).toBeDefined()
+    })
+
+    it('should get the batch', async () => {
+      const data = generateDummyBatchForTesting()
+      const result = await provider.getOne<Batch>(R_BATCHES, { id })
+      const { id: omittedId, ...expectedResult } = result.data
+      expect(expectedResult).toMatchObject(data)
```
IanMayo, line 54: note: I don't think we should be omitting the `id` when testing `getOne`. It should be the `id` we used for `getOne`.

IanMayo:
Thanks for the new approach @TahaKhanAbdalli 

I want to give us a really good foundation for these data tests - so I'm going to carry on reviewing it on Monday.

IanMayo:
Hello @TahaKhanAbdalli - we don't use the password declared in the `.env` file.  I guess we can delete that file, and the `jest` config that loads the file.

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  describe('CRUD Operation on Batch', () => {
+    it('should create a batch', async () => {
+      const result = await provider.create<Batch>(R_BATCHES, {
+        data: generateDummyBatchForTesting()
+      })
+      id = result.data.id
+      batch = result.data
+      expect(id).toBeDefined()
+      expect(batch).toBeDefined()
```
IanMayo, line 61: IMHO this test should also `get` the record from the provider as a way of verifying it got created.

Here we're just verifying the object that got returned.

Oh, and before we `create` the record, can we run a test to verify the data-provider is empty?  (`getList` should allow a count)

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    })
+
+    it('should get the batch', async () => {
```
IanMayo, line 64: If we test `get` in `create` we don't need this test, IMHO.

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,108 @@
...
+          createdAt: DateTime.now().toFormat('yyyy-MM-dd')
+        }
+      })
+      const shouldMatchBatch = generateDummyBatchForTesting({
+        name: 'dummy-batch',
+        id
+      })
+      const expectedResult = result.data
+      expect(expectedResult).toMatchObject(shouldMatchBatch)
+    })
```
IanMayo, line 96: again - here we are testing the object that got returned. We aren't testing that the entry in the data-store got modified.  We also need to do a `get`.

src/providers/dataProvider/dataprovider.test.ts
IanMayo: The last failing `get` has been lost. Can you put it back please?

src/providers/dataProvider/dataprovider.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    receiptNotes: 'receipt-notes-1',
+    createdAt: DateTime.now().toFormat('yyyy-MM-dd'),
+    createdBy: 1
+  }
+}
+
+describe('CRUD operations on each resource', () => {
+  let provider: DataProvider
+  let id: number
+  let batch: Batch
```
IanMayo, line 41: @TahaKhanAbdalli - I'd like all of these tests to be independent.  I think we can _guarantee_ that by re-creating the data-provider in the `beforeEach`, rather than the `beforeAll`.

IanMayo:
I've _tried_ to help out, by doing some refactoring.  But, see note [here](https://github.com/DeepBlueCLtd/RCO/pull/236/commits/b9aaf31ff69e405b4349ba6bc73be8ee7e0ef44e#diff-e7a5ae082551db708f01c0f0e1146aa569273bd9bf4b4db4b0fe5d26aad53512R46) my J/S skills aren't up to refactoring code with `await` falls into a separate method.

IanMayo:
I've made some changes @TahaKhanAbdalli - to do more testing of the audit log.

But, since merging the PR for `audit` log we have some conflicts.  Would you mind trying to solve them please?  I think the root of the problem is the refactoring you had to do to get at the `lifecycleCallbacks.  I _guess_ it's worth trying to re-apply your changes to the current version of `DataProvider.index` in `main`.

IanMayo (approved):

IanMayo (approved):

PR #237, opened 2023-05-05
Loan Process - 2nd version

### LINKED ISSUE #232: Loan Process - 2nd version
My initial perceptions were that a Loan was for one User for one Batch.

But a Loan is for all items lent from a batch.

So, `Loan` doesn't actually store any unique data - it's just a list of items from a batch.  

We can _actually_ support that by just having a `loanedTo` field in `item`:
```
loanedTo: User['id''] | undefined
```

Note: this strategy removes the `Loan` and `LoanItem` tables, and makes it **much** easier find items loaned to a user.

So, when a group of items are loaned to a user, we just put that userId in `loanedTo`.

When items are returned, we just clear the `loanedTo` flag.

For the `Loan` process we have a remark field.  If it is provided, it will get included in the audit entry.

We need a bit of an audit trail to cover this, so for the loan, we create an audit entry for:
```
activity: `ITEM_LOAN`
comment: `Item loaned to ${user} by ${currentUser}. ${remark}`
id: itemId
```

```
activity: `ITEM_RETURN`
comment: `Item returned to ${user} by ${currentUser}. ${remark}`
id: itemId
```

Note: for the above we need to introduce an `id` field for `trackEvent`.  This will be eventually id for a number of table types, but we know which from the `activity` type.

Note 2: the `item` will still have a value in `vaultLocation`. The users actually want this - so that when the item is returned, they know which shelf to put it back on.  But, we'll have to modify the `Muster Reports` to only include items that are not on loan.

Note 3: with this new data model in place we can add `Loaned to` to the `ItemList` filters and columns (hidden by default).

salony75way:
fixes #232


IanMayo:
Compile error to be fixed, please @salony75way :

Error: src/providers/dataProvider/index.ts(83,11): error TS2339: Property 'name' does not exist on type 'true | { name: string; }'.

IanMayo:
Hello @salony75way - the architecture still isn't quite right. We no longer require the `Loan` entity.  We just have some loan-related attributes to `Item`.   Updated schema [here](https://docs.google.com/drawings/d/1SYIb0PZxcabvhXdNMAEHlbdWaoam2ipB7x5ZUudVVGc/edit)

salony75way:
> Hello @salony75way - the architecture still isn't quite right. We no longer require the `Loan` entity. We just have some loan-related attributes to `Item`. Updated schema [here](https://docs.google.com/drawings/d/1SYIb0PZxcabvhXdNMAEHlbdWaoam2ipB7x5ZUudVVGc/edit)

Hi @IanMayo , Yes i have removed all resources related to loan. Let me know if we need to remove some other dependency in code.

IanMayo:
> Hi @IanMayo , Yes i have removed all resources related to loan. 

We still have the `Loan` type, and we still refer to it from `LoanForm`:
image

IanMayo:
One other change please. It's about `loanedBy`.  This suggests it is the librarian that lent the book. This should be `loanedTo` - so it's the person that has taken the book away.

IanMayo:
Hello @salony75way - IMHO the `loan` form and buttons should be moved to the `Items` resource - since they are only used for that resource.

IanMayo:
Aah, one more thing. Can we have a `notify` popup once the database changes are complete?

src/providers/dataProvider/index.ts
```diff
@@ -78,33 +78,76 @@ const withCreatedBy = (
...
+        data: { name }
+      } = await provider.getOne<User>(constants.R_USERS, {
+        id: holder
+      })
+
+      const promisees = items.map(async (item) => {
+        await audit({
+          type: AuditType.ITEM_LOAN,
+          activityDetail: `Item loaned to ${name} by ${userName}. ${remarks}`,
+          resource: constants.R_LOAN_ITEMS,
```
IanMayo, line 106: This should be `constants.R_ITEMS`

We should probably delete `constants.R_LOAN_ITEMS`

src/types.d.ts
```diff
@@ -108,6 +103,7 @@ interface Item {
   musterRemarks: string
   createdAt: string
   createdBy: User['id']
+  loanedBy?: number
```
IanMayo, line 106: this should be `loanedTo`

IanMayo:
Have just been talking to users. They have said they don't need the `Remarks` field in the `LoanForm`.   It's not on the legacy system, nor is it in the system their sister-organisation uses.

So, it will get deleted from the `remarks` too, please.

PR #242, opened 2023-05-05
audit for data item update

### LINKED ISSUE #230: Log item edits
For audit purposes we wish to log all item edits.

The `trackEvent` should gain an optional `index` parameter, that contains the record it.  

So, for each `Update` we should perform the following:
- compare `record.previousData` with `record.data`
- if there are changes, modify the `audit` method call to include a series of `label`: `previousValue` pairs showing the previous values of modified fields.
- move the record id to the `index` parameter

Note: I _think_ we need to use `beforeUpdate` if we want to see the `previous` and `current` values of the object.

Hopefully we can find a lib that does the diff for us.

Note: we shouldn't do this for `audit` items, since that would become recursive. Aah, hold on: we don't have a `DataProvider` handler for audits, so that's ok.

salony75way:
fixes #230


IanMayo:
I've just played with this @salony75way 

1. I changed the end date for an item, but it wasn't logged
2. The audit entry should include the changes in the `Remark` but instead they are stored in a `previousValue` field which isn't shown in the UI.  I'd like the changes to be listed in field/value pairs in the Remark, please.

We already store the `user` that made the change.  So I guess the remark just needs to be:
```
Made edits to this item. Previous values: ${list-of-key-value-pairs}`
```

salony75way:
> We already store the user that made the change, and we know that it's an item_edit activity, so I guess the remark just needs to be:

Hi @IanMayo, Can you please elaborate a bit more this?

IanMayo:
> > We already store the user that made the change, and we know that it's an item_edit activity, so I guess the remark just needs to be:
> 
> Hi @IanMayo, Can you please elaborate a bit more this?

Sure.  For some of the detail in the `remark`, we know that information from other columns (who is doing it: `ian`, and what the change is: `item_edit`).  So, for an edit, I suggest the `remark` column just needs to store the previous label values of anything that changed.  We don't need to store what the value was changed to, since that is shown in the current fields of the `item`.

So, if I change the `protective marking` for an `item` from `prot-mark-3` to `prot-mark-5` and the media-type from `CD` to `Tape`, then the remark in the `Audit` for that change will be `Previous values: Protective Marking: prot-mark-3, Media Type: CD`.

salony75way:
> 

Hi @IanMayo, ATM we don't have any `remarks` field for audit, So let me know if i need to add this or you are referring the `activityDetail` field here?

IanMayo:
> > 
> 
> Hi @IanMayo, ATM we don't have any `remarks` field for audit, So let me know if i need to add this or you are referring the `activityDetail` field here?

Yes, `activityDetail ` please.

IanMayo:
Aah, we need to remove `activityDetail` and store the data in `label` instead, please.

salony75way:
> data_id

Hi @IanMayo , Do i need to remove `activityDetail` from audit schema  or i need to remove just for items?

IanMayo:
let's mark it as deprecated in the schema please. When there are no open PRs I'll delete it.

IanMayo (approved):

PR #243, opened 2023-05-08
234 audit history for items

### LINKED ISSUE #234: View audit history for items
For `ItemShow` we should introduce a history tab - which shows all audit events for that item (`AuditList` component).

It will be fairly expensive to get the data for the tab, so only populate it when the tab is opened (or at least get the data once the ItemShow component has finished loading).

This list will contain all the `audit` events for this item.

We have separate work to introduce the `id` into the audit log. But, for this history to be generated, I guess we need to add an (optional) `resource` field to the `trackEvent` method, and the audit resource. This will be one of `item`, `batch`, `project`, etc. etc.

So, to get the history for the current item we filter the `AuditList` component for `resource: item` and `id: item.id`.

We will be able to filter the audit history by:
- created
- activity type
- user
- security related

TahaKhanAbdalli:
Fixes #234


IanMayo:
Thanks. For User - can we show the user name please?
image

TahaKhanAbdalli:
Should this also reflect in `audit list` in `reference data`, because now only the id is showing?

IanMayo:
> Should this also reflect in `audit list` in `reference data`, because now only the id is showing?

Yes, it would be useful to include these filters, please.

IanMayo (approved):

IanMayo:
Oh dear, I've got myself stuck here @TahaKhanAbdalli 

I had to change the `AuditList` to filter on `data_id` rather than `id` (since we want to filter for audit entries that relate to this item - not for the single audit entry that has the same id as the current item).

I did that with [this change](https://github.com/DeepBlueCLtd/RCO/pull/243/commits/e59ae1a128c7324c2cd65508436ae7bfc97fce5f#diff-a4c81cf46fb6e54fb36ef8f0857e8271b1c80c7e7358bf934fb2384ceca43c47R26). But, the compiler doesn't like it. Can you help out please?

PR #244, opened 2023-05-08
Improve date search for items#79

### LINKED ISSUE #79: Improve date search for items
The search fields for start/end date for items don't work in a useful way.

~~We need to let users choose to search by either start or end date, and to specify whether they are looking for items before or after the date specified.~~

Hmm, I think we need something more complex for this.  I _think_ we need a third-party DateRange input component. The user specifies the range of time for which they want items with data, then we filter in this way:
```
- item.start < filter.end
AND
- item.end > filter.start
```

This will give items that have data:
- of greater time range than the filter range
- with time range completely included in the filter range
- with time range overlapping the filter range

The search control for these fields will not initially require time to be provided. But, let's try to use a component that could allow time to be provided in the future.

salony75way:
fixes #79


IanMayo:
I know this isn't complete yet. I've just had a play:
- searching didn't work. I could see an item that was `live` from 3rd to 4th October 2020.  I provided the whole of October as the filter period, but no data matched.  
- I removed the DataRange filter, and no records were shown.  From the `Add Filter` menu I clicked on `Remove all filters`, and all data was returned.

src/providers/dataProvider/index.ts
```diff
@@ -300,6 +300,8 @@ export const getDataProvider = async (
     {
       resource: constants.R_ITEMS,
       beforeCreate: async (record: CreateResult<Item>) => {
+        record.data.start = new Date(record.data.start).toISOString()
```
IanMayo, line 303: hello @salony75way - wouldn't it be more effective to change the data generation so that the start and end values are expressed as ISO strings?

IanMayo:
Hello @salony75way - you approach is logical, but it doesn't quite give me what we need.  The current filters will find items that start after the start date, and end before the finish date.  This is logical, but it only gives items that are completely within the time range.

I actually want items that span the time range.  So, where the requested range is completely within the item range, or where the item range overlaps with the requested range.

To get this, we need to filter for items where `item.start` < `filter.end` AND `item.end` > `filter.start`.

src/providers/dataProvider/index.ts
salony75way: Hi @IanMayo, Filter was not working with date string so i just changed the date to ISO string format

IanMayo: Yes - but let's correct it where we first generate the start/end dates.  Then we won't have to convert it here - or anywhere else.
We _may_ wish to cosmetically `format` dates for display, but it makes sense to store them internally as ISO dates.

IanMayo:
Hello @salony75way - this still doesn't seem right, I'm afraid.  I'd love to step in and sort it, honest - I just don't (yet) understand react-admin filtering.

Here is some sample data:
Screenshot 2023-05-10 at 08 58 13

Note that item 2 goes from the 7th to 13th of Nov 2022.    If I apply a filter for 8th to 10th Nov 2022 I should get that item, but I don't:
Screenshot 2023-05-10 at 08 58 00

Here is the logic: 
> To get this, we need to filter for items where item.start < filter.end AND item.end > filter.start.

So, in the above example the item start date is < the filter end date AND the item end date is after the filter start date - so it should be matched.

IanMayo (approved):

PR #245, opened 2023-05-08
Loan and loan returns for relevant items

### LINKED ISSUE #238: Only enable `Loan` and `Loan Return` for relevant items
For `ItemList` we have `Loan` and `Return` buttons for bulk selection.

We know the `Loaned` state for an item is `loanedTo` is non-null.

`Loan` should only be enabled where the `None` of the selected items are loaned 
`Return` should only be enabled where the `All` of the selected items are loaned 

Note: this can't be done until #232 is complete.

TahaKhanAbdalli:
Fixes #238

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -23,7 +23,7 @@ import * as constants from '../../constants'
 import CreatedByMeFilter from '../../components/CreatedByMeFilter'
 import { ItemAssetReport } from './ItemsReport'
 import { Button, Modal } from '@mui/material'
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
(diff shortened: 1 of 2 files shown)


IanMayo (approved):

TahaKhanAbdalli:
> Nice code @TahaKhanAbdalli - but a couple of minor observations.
> 
> 1. If `selectedItems` is of zero length we don't need to run the filter
> 2. I _think_ we can use `useState` to only re-run the function if `data` or `selectedIds` changes.  I've [had a go](https://github.com/DeepBlueCLtd/RCO/pull/245/commits/22289bb593bc9709c0af6d7f019103615e0c22d9) at this change, just to check that we do reduce re-calculation.  Could you inspect the code please, and verify it's ok?

Yes the code is ok

PR #246, opened 2023-05-08
Muster List for users#136

### LINKED ISSUE #136: `Muster List` for users
We have the `Users` list in `Reference Data`.  

Users sometimes need to find a list of all of the `items` for a `User`.

So, from the `Users` page we should be able to select a user and then click on `Muster List` from the toolbar.

This will create a list of all of the items loaned to that `user`.  This will be formatted similar to the report we produce from `BatchShow`.

The list will have these columns:
- item number
- media type
- protective marking

Note: we can't start this until we have `Loans` in application

Some new requirements:
1. We need the two-column signature block at the bottom, like we have for `Vault Location Muster List`.
2. We show these columns in the table:
    a. Item Ref Number
    b. Composite of `Project` & `Platform` fields (taken from parent `Batch`).
    c. Date loaned (we'll have to extract this from the audit log for the item - it's data of the latest `Loan` activity.
    d. Media Type
    e. Prot Marking
    f. No. Pages / Serial Number (sorry - this is still missing from the data model)
3. Header line which reports number of items on loan to that user `${num} Items Currently Booked Out`
4. Header line which includes current date and time.
5. Title line which says "RCO - Loans to ${name}`

salony75way:
fixes #136


IanMayo:
Hi, have added some new requirements, please: https://github.com/DeepBlueCLtd/RCO/issues/136

IanMayo:
One extra feature, please @salony75way   The `loaned date` in the form is always empty.  I think we need to populate that value in the `loan` method. Could you make that change please?

IanMayo (approved):

PR #253, opened 2023-05-10
refactor item show

### LINKED ISSUE #250: Refactor ItemShow
We are going to refactor the `itemShow` component.

We now no longer require the media tabs.

We are going to add history for that item to the right-hand side of the component, as shown here:
https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g221ed9988a1_0_57

TahaKhanAbdalli:
Fixes #250

### DIFF
--- a/src/resources/items/ItemShow.tsx
+++ b/src/resources/items/ItemShow.tsx
@@ -4,16 +4,16 @@ import {
   Form,
   Loading,
   Show,
-  TabbedShowLayout,
   TopToolbar,
   useShowContext
 } from 'react-admin'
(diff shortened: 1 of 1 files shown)


IanMayo (approved):
If possible, could you fix the left-hand side to 550px wide, and let the history consume the remaining horizontal space?

IanMayo (changes requested):
If possible, could you fix the left-hand side to 550px wide, and let the history consume the remaining horizontal space?

PR #259, opened 2023-05-10
mangle login page, so e2e fails

IanMayo:
do not merge. This is just re-testing e2e for login page


IanMayo:
Close. Handled in alternate PR #264

PR #263, opened 2023-05-11
Multiple selection filter for activity type field

### LINKED ISSUE #256: Multi-select activities from filters for Audit List
In the Audit List, users can select the type of activity to filter for.

But, we can only select one item

I'd like to be able to select multiple items from this list. For example, so I can show both `Item Loan` and `Item Return`.

salony75way:
fixes #256


IanMayo (approved):
Wow, that was easy!

PR #265, opened 2023-05-11
User permissions and roles

### LINKED ISSUE #173: Permissions
We (currently) have login via username and password.

If it only applies to a limited number of users, this isn't a burden.

I do believe it is possible to have a workable system where most users don't need to login - removing the burden of integrating some external authorisation system, or having another username/password for all staff.

_Eventually_ the user roles will come in with the HTTP properties.  As a substitute for that, we will add roles to our `users` table.

We are going to have a `role-based` permission model.  We will have these two roles:
- `rco-user`
- `rco-power-user`

We will modify the `users` table to have a `roles` field. This will contain an array of strings.  The `UserList` form will show these strings in a series of lozenges/badges.  The `UserEdit` form will use a multi-select drop-down to allow these string items to be edited.

Note: when we make this change, we will remove the `admin` property from `User`.

## Users (no authentication)
- view lists of `projects`, `batches`, `items` (including item loan state)
- ability to filter data by all possible fields
- ability to control visible columns in all list views
- ability to (eventually) submit a request from the `Items` page relating to a selected set of items
- view list of users
- create a muster list for a user

## `rco-user` (username/password)
- as above, plus:
- add/edit `projects`, `batches`, `items`, `platforms`
- loan/return items
- create destruction certificate for items
- despatch items
- change item locations
- create muster forms

## `rco-power-user` (username/password)
- as above, plus:
- delete `projects`, `batches`, `items`
- add/edit/delete other `ReferenceData` tables, notably including:
    - users
    - protective marking lists

salony75way:
Supports #173


src/providers/dataProvider/defaults/users.ts
```diff
@@ -4,28 +4,32 @@ const users: User[] = [
...
+    roles: ['rco-power-user']
   },
   {
     id: 2,
     name: 'jason',
     password: 'user',
     adminRights: false,
-    active: true
+    active: true,
+    roles: ['rco-power-user']
```
IanMayo, line 16: could you make `jason` the plain `rco-user` please. So we then have one active user for each of the roles.

IanMayo:
@salony75way - it _looks-like_ I don't get a deployment to review until you take it out of `Draft`.  I'll be able to give some UI feedback if you take it out of review.

(don't worry - I know you still have a long way to go :-) ).

src/providers/dataProvider/defaults/users.ts
salony75way: Done

IanMayo:
Have made a couple of edits to the original issue, but they don't make a great difference:
image

src/types.d.ts
```diff
@@ -5,13 +5,16 @@ interface CustomDataProvider {
 
 type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
 
+type UserRole = 'rco-user' | 'rco-power-user'
+
```
IanMayo, line 9: @salony75way - we have `rco-user` and `rco-power-user` written as strings in two different files.  As soon as I see this, I think _these should go into the constants files_ :-)

IanMayo:
Hey @salony75way - it's really impressive how small the react-admin changes are to introduce this list of roles :-D

src/types.d.ts
salony75way: @IanMayo  This is the type file for user roles

IanMayo: Aah yes, my mistake.  I was worried that we may be able to change `rco-power-user` to `rco-super-user` above, and the change not be spotted in `options.ts`.  But, I see in `options.ts` that it is declared of type `UserRole` - so the compiler would spot the error.

IanMayo:
@salony75way - this introduces the new roles, but doesn't implement the `permissions` (controlling access to pages) in the UI.  Shall we merge this, and implement the `permissions` in another PR?

salony75way:
> @salony75way - this introduces the new roles, but doesn't implement the `permissions` (controlling access to pages) in the UI. Shall we merge this, and implement the `permissions` in another PR?

Yes we can merge this

IanMayo (approved):

PR #266, opened 2023-05-12
refactored data provider

### LINKED ISSUE #262: Refactor `DataProvider`
I was investigating how to refactor `DataProvider` - since IMHO it's too long for one module.

I was considering moving resource handlers to separate files.

But, I now see there is a lot of duplication of boilerplate code in the `DataProvider\index` file.

To this, I propose we create this method: 
* `auditEntry = (type: AuditType, securityRelated: boolean, resource?: string, id?: number)`

TahaKhanAbdalli:
Fixes #262


src/providers/dataProvider/dataprovider-utils.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (s1 < s2) {
+    return -1
+  } else if (s1 > s2) {
+    return 1
+  } else {
+    return 0
+  }
+}
+
+export const generateBatchId = async (
```
IanMayo, line 26: Since this method is only called from `Batch`, can we put it in the Batch file?

IanMayo (approved):

PR #268, opened 2023-05-12
item resource tests

TahaKhanAbdalli:
Supports #180


src/providers/dataProvider/tests/item.test.ts
```diff
@@ -0,0 +1,276 @@
...
+      prefixLocalForageKey: TEST_STORAGE_KEY
+    })
+    provider = withLifecycleCallbacks(
+      withOutLifecycleProvider,
+      lifecycleCallbacks(
+        trackEvent(withOutLifecycleProvider),
+        withOutLifecycleProvider
+      )
+    )
+
```
IanMayo, line 53: @taha - there are two very similar blocks in the next few lines.

I think we should declare an array of resource-names, then loop through the names clearing each.  We should use the same strategy in batch-test.

We will have tests for other resource types in the future, so we will just end up passing in the full list of resource names.

taha: You have the wrong Taha there 😅

src/providers/dataProvider/tests/batch.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      pagination: { page: 1, perPage: 1000 },
+      filter: {}
+    })
+    if (list.total !== undefined && list.total > 0) {
+      await provider.deleteMany(R_BATCHES, {
+        ids: list.data.map((item) => item.id)
+      })
+    }
+
+    const auditList = await provider.getList<Audit>(R_AUDIT, {
```
IanMayo, line 57: @TahaKhanAbdalli - we should be using your generic clear method here in this class.

src/providers/dataProvider/index.ts
```diff
@@ -216,7 +216,7 @@ const auditForUpdatedChanges = async (
     activityDetail: `Previous values: ${JSON.stringify(difference)}`,
     resource: constants.R_ITEMS,
     index: record.id as number,
```
IanMayo, line 218: @TahaKhanAbdalli  - I don't see why we are using `previousData` in the next line.  Are the two ids always the same?

TahaKhanAbdalli: Yes the `ids` are same, when updating `react-admin` provides the full current document to update in `record.data` but when we are manually updating in the test we provided `partial` of the data to update which was causing issues in testing. Or I can remove this and provide the id when updating the document then we can use `record.data.id`

IanMayo: Aah, I see.  That's ok :-)

IanMayo (approved):

PR #270, opened 2023-05-12
Added permissions by roles

### LINKED ISSUE #173: Permissions

salony75way:
fixes #173

### DIFF
--- a/src/providers/authProvider/permissions.ts
+++ b/src/providers/authProvider/permissions.ts
@@ -1,0 +1,68 @@
+type UserRoles = UserRole | 'user'
+
+const permissions: Record<UserRoles, ResourcePermissions> = {
+  user: {
+    projects: { read: true, write: false, delete: false },
+    batches: { read: true, write: false, delete: false },
+    items: { read: true, write: false, delete: false },
+    users: { read: true, write: false, delete: false }
+  },
+  'rco-user': {
+    projects: { read: true, write: true, delete: false },
+    batches: { read: true, write: true, delete: false },
+    items: { read: true, write: true, delete: false },
+    users: { read: true, write: true, delete: false },
+    platforms: { read: true, write: true, delete: false }
+  },
+  'rco-power-user': {
+    '*': { all: '*' }
+  }
+}
+
+export default permissions
+
+const generatePermission = (
+  permission1: Permission,
+  permission2?: Permission
+): Permission => {
+  if (permission2 == null) return permission1
+
+  if (permission2.all === '*' || permission1.all === '*') {
+    return { read: true, write: true, delete: true }
+  }
+
+  const read =
+    typeof permission1.read !== 'undefined' && permission1.read
+      ? true
+      : permission2.read
+  const write =
+    typeof permission1.write !== 'undefined' && permission1.write
+      ? true
+      : permission2.write
+  const deletePermission =
+    typeof permission1.delete !== 'undefined' && permission1.read === true
+      ? true
+      : permission2.delete
+
+  return { read, write, delete: deletePermission }
+}
+
+function mergedPermission(
+  permissions: ResourcePermissions[]
+): ResourcePermissions {
+  const results: ResourcePermissions = permissions[0]
+  permissions.slice(1).forEach((permission) => {
+    Object.entries(permission).forEach(([key, value]) => {
+      results[key] = generatePermission(value, results[key])
+    })
+  })
+  return results
+}
+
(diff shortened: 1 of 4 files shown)


src/providers/authProvider/permissions.ts
```diff
@@ -0,0 +1,68 @@
+type UserRoles = UserRole | 'user'
```
IanMayo, line 1: This is an impressive module @salony75way - you've clearly done this before 🤣 

I wouldn't have had a clue where to start ;-)

IanMayo:
I've had a play @salony75way . Two changes please:
1.  to start off with, all users should be able to read all reference data.
2. we need to allow `unregistered` users to browse the data without logging in. So, they won't have user accounts.

src/providers/authProvider/permissions.ts
salony75way: @IanMayo  Yes i did this once, But in react admin we have Role-Based Access Control (RBAC) library but its only in  enterprise edition so we have to create it custom.

IanMayo:
Thanks @salony75way .  I can't "create" items as either `rco-user` or `power-user`.

I _think_ we should  introduce `create` permission, in addition to `write`.

IanMayo:
Hello @salony75way - while it appeared broken, after a rebuild, followed by logout/login, it seems to be working.

But, IMHO we need some brief unit tests for these methods:
- `protectedRoutes`
- `useCanAccess`

That will make it easier to test edge cases, and spot anything that may break in the future.

salony75way:
> 

As of now we are including `update` permission in `write`, So please let me know if we need only `create` permission here or i need to introduce a new permission (`update`)?

IanMayo:
> As of now we are including `update` permission in `write`, So please let me know if we need only `create` permission here or i need to introduce a new permission (`update`)?

Hmm, I've been having a play.  The current levels of permission do seem ok. So no changes to levels of permission :-)

src/providers/authProvider/permissions.ts
```diff
@@ -0,0 +1,111 @@
...
+    platforms: { read: true, write: true, delete: false },
+    'reference-data': { read: true, write: true, delete: false }
+  },
+  'rco-power-user': {
+    '*': { all: '*' }
+  }
+}
+
+export default permissions
+
```
IanMayo, line 24: @salony75way - for the following method, I wish to understand the two parameters.

Is the first parameter the set of permissions being requested, and the second parameter the responses to those requests?

IanMayo:
One small change please. Could we only who the `Edit` button for a Show form if the user has write permission?  Currently it jumps to the relevant `List` view, but IMHO it would be better if the Edit button was either deleted or disabled if the user doesn't have `write` permission for that resource.

src/providers/authProvider/permissions.ts
salony75way: Hi @IanMayo,
`generatePermission` method is a helper for `mergedPermission` to merge permission if user has multiple permissions.

`generatePermission` method takes two arguments, first is permission for resource  and second is merged permissions object.

IanMayo (approved):

PR #276, opened 2023-05-15
removed name of the resource from activity type

### LINKED ISSUE #274: Refactor - remove  resource name from ActivityType for resources
In our audit log we have an activity type, with entries like "edit_item" and "create_platform".

Since we have an optional `resource` column in the audit log, we can remove the name of the resource from the enumerated type.

Activities that don't relate to resources (login, logout) would not change.

This will allow the users to filter for all `Create` activities.

TahaKhanAbdalli:
Fixes #274

### DIFF
--- a/src/utils/activity-types.ts
+++ b/src/utils/activity-types.ts
@@ -2,26 +2,26 @@ export enum AuditType {
   LOGIN = 'login',
   LOGOUT = 'logout',
 
-  CREATE_USER = 'create_user',
-  DELETE_USER = 'delete_user',
-  EDIT_USER = 'edit_user',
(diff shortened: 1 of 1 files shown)


IanMayo:
This partially fixes the issue. But, I think we can take advantage of the simplification if we reduce the constants to just `CREATE`, `EDIT`, and `DELETE`.

I think this will allow us to refactor some DataProvider code, in the longer term.

IanMayo (changes requested):
Thanks @TahaKhanAbdalli, but we need to audit platform-create please.

IanMayo (approved):

PR #289, opened 2023-05-16
promoted users to left hand menu and versioning buf fixed

### LINKED ISSUE #284: Promote "Users" to left-hand menu
Lots of business in this app will be related to returning items.

The start point of this process is the user returning the items.

So, we should add a `Users` button on the left-hand menu.

We can remove it from the `Reference Data` tab.

TahaKhanAbdalli:
Fixes #284

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,7 +9,7 @@ import {
 import { Route } from 'react-router-dom'
 import MyLayout from './components/Layout'
 import React, { Suspense, useEffect, useState } from 'react'
-import { AllInbox } from '@mui/icons-material'
+import { AllInbox, Groups } from '@mui/icons-material'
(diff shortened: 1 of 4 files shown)


IanMayo (approved):

IanMayo:
> versioning bug fixed 

@TahaKhanAbdalli - can you explain what the versioning bug was please?

TahaKhanAbdalli:
In the `.env` the version was 2 but in my `localstorage` it was 1 so every time I reloaded the page it was loading the `default data`. In the `if` statement it was just comparing the versions and if they didn't match it will load defualt data however, after loading default data it should be putting the version along with it in `localstorage`.

IanMayo (approved):

PR #291, opened 2023-05-16
User show page created and departed logic implemented

### LINKED ISSUE #281: Introduce departedDate for users
Staff have to go through a process when a member of staff is about to leave.

They have to verify that person has no items on loan, and then store the departure data.

So, add a `Depart Organisation` button to the `UserShow` form.  The button is only enabled if the user has no items on loan.

Aah. we don't currently have `UserShow`.  So, clicking on item in `UserList` should open `UserShow`, which shows the `UserForm` (read-only) on the left-hand side, and a `ItemList` of items loaned to that user on the right-hand side.  Staff should be able to `Return` items from that `UserList`.

When the button is clicked, mark the user as `not active`, store the current date in the `departedDate`, and store `USER_DEPARTED` in the audit log (security-related change).

There will need to be an `Edit` button on `UserShow`, to get to `UserEdit`.

TahaKhanAbdalli:
Fixes #281


IanMayo:
Thanks. Will hold this until we've merged `Permissions`.  It may be easier for you to re-apply your changes to `develop` than to re-apply the permissions changes.

IanMayo:
@TahaKhanAbdalli - are you able to resolve the conflicts in this PR please?

PR #296, opened 2023-05-17
Added consecPages field to items

### LINKED ISSUE #51: Introduce item descriptor (with explanations)
**Update: 03/05/2023** - the RCO replacement can get buy with a simpler concept of media-type.

We no longer need the media tabs.  Nor do we require the media-specific objects in `Item` - the types can be deleted.

We have these concepts:
1. LEP (a list of effective pages document is present in the printed media)
2. Number of sheets
3. CONSEC number (consecutive counter across a batch of items)
4. N/A - for mag media without consec.

Items 1 and 2 are for printed media, and are mutually exclusive.  Item 3 is relevant for some types of mag/optical media, but not others (4).

Ok. we will just add a single field to `Item`.  It will be called `consecPages`.  It will store one of the above fields.  But a text box next to it will describe what to put in it

Note: once we have the `consecPages` field in `Item`, we should add the value to the `Vault Location Report`, `User Muster List` and `Asset Report` (from `ItemList`).  The header label should be `Serial/Pages`.

Note: I'll update the UX to show this. We'll add text box for `consecPages`, with a large text label next to it.  I'll fill in the text for the label.

image

salony75way:
fixes #51


src/resources/items/ItemForm/index.tsx
```diff
@@ -116,24 +99,13 @@ export default function ItemForm({ isEdit }: FormProps): React.ReactElement {
...
-          <MediaForm type='Tape' source='magTape' />
-        </TabbedForm.Tab>
-        <TabbedForm.Tab label='DVD' icon={<Album />}>
-          <MediaForm type='DVD' source='dvd' />
-        </TabbedForm.Tab>
-        <TabbedForm.Tab label='Paper' icon={<MenuBook />}>
-          <MediaForm type='Paper' source='paper' />
-        </TabbedForm.Tab>
-      </TabbedForm>
+        <CoreForm batchId={batch?.id} />
```
IanMayo, line 107: now that we have removed the tabs, do we still need the separate CoreForm?

Would it be tidier to bring the component into this module?

IanMayo (commented):
@salony75way not at PC, but does the UI include the label for me to explain ConSec?

src/resources/items/ItemForm/index.tsx
salony75way: No, I will move the code to ItemForm component. Thanks!

salony75way: @IanMayo  
We already using CoreForm in ItemShow component so there will be more handling in ItemForm component if we will merge both CoreForm to ItemForm components. Let me know if we still need to do this?

salony75way:
> @salony75way not at PC, but does the UI include the label for me to explain ConSec?
Its look like this:

Screenshot 2023-05-17 at 2 30 47 PM

src/resources/items/ItemForm/index.tsx
IanMayo: Oh, if it's used in two places, let's keep it as a component.

IanMayo (approved):

PR #299, opened 2023-05-17
Destroy items

### LINKED ISSUE #269: Destroy Items
When an asset is no longer required, users will choose to destroy them.

This is a tightly controlled process, involving a destruction certificate.

The process and UI for destruction is recorded [here](https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g2419163b40d_0_15)

The form to display on `Print` is displayed [here](https://docs.google.com/document/d/1qYGtDE1_wUXtkWcoqxh9AwMUMe7Rzs3s5zxJuNBgzyY/edit?usp=sharing)

salony75way:
fixes #269

### DIFF
--- a/src/resources/items/DestroyItems.tsx
+++ b/src/resources/items/DestroyItems.tsx
@@ -1,0 +1,80 @@
+import { Box, Button, Typography } from '@mui/material'
+import FlexBox from '../../components/FlexBox'
+import { useDataProvider, useNotify } from 'react-admin'
+import * as constants from '../../constants'
+import React, { useState } from 'react'
+import Printable from '../../components/Printable'
+
+interface Props {
+  ids: number[]
+  onClose: () => void
+  successCallback: () => void
+}
+
+const style = {
+  position: 'absolute' as 'absolute',
+  top: '50%',
+  left: '50%',
+  transform: 'translate(-50%, -50%)',
+  width: 400,
+  bgcolor: 'background.paper',
+  border: '2px solid #000',
+  boxShadow: 24,
+  p: 4
+}
+
+export default function DestroyItems(props: Props): React.ReactElement {
+  const { onClose, successCallback, ids } = props
+  const [open, setOpen] = useState(false)
+  const datProvider = useDataProvider()
+  const notify = useNotify()
+
+  const onDestroy = (): void => {
+    datProvider
+      .deleteMany(constants.R_ITEMS, { ids })
+      .then(() => {
+        notify(`${ids.length} items destroyed!` )
+        setOpen(true)
+        successCallback()
+      })
+      .catch(console.log)
+      .finally(() => { setOpen(true) })
+  }
+
+  return (
+    <Box sx={style}>
+      <Typography variant='h6'>Destroy {ids.length} items:</Typography>
+      <FlexBox marginTop={3}>
+        <Button variant='contained' onClick={onDestroy}>
+          Destroy
+        </Button>
+        <Button color='secondary' variant='outlined' onClick={onClose}>
+          Cancel
+        </Button>
+      </FlexBox>
+      <DestroyItemsReport
+        ids={ids}
+        open={open}
+        onClose={() => { setOpen(false) }}
+      />
+    </Box>
+  )
+}
+
+interface DestroyItemsReportProps {
+  ids: number[]
+  open: boolean
+  onClose: () => void
+}
+
+const DestroyItemsReport = (
+  props: DestroyItemsReportProps
+): React.ReactElement => {
+  const { open, onClose } = props
+
+  return (
+    <Printable open={open} onClose={onClose}>
+      <Box>TEST</Box>
+    </Printable>
+  )
+}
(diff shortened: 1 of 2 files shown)


IanMayo:
Could we have a `Remarks` box in the `DestructionCreate` form, please?

IanMayo:
* [ ] We should not be able to `Destroy` if an item is on loan
* [ ] DestructCreate has `CRATE` (not `CREATE`) as the command button
* [ ] On `Destroy` in ItemList, if no live destructions, show message saying "Please create the destruction job before destroying items".
* [ ] On `Destroy` in ItemListthe popup title should be `Add xx items to destruction job`
* [ ] On `Destroy` in ItemList in the popup, the drop-down list title should be `Destruction Jobs`
* [ ] For `DestructionShow` show table on right-hand side that lists the items in this destruction
* [ ]  For `Destroy` on `DestructionShow`, show confirmation after update complete.

IanMayo:
* [ ] The confirmation from `ItemList` should be `3 items added to Destruction Job` (because they haven't been destroyed yet).
* [ ] `DestructionList` should have filters for `Pending Jobs` and `Completed Jobs`. These filter on `finalisedAt` = or != null respectively.
* [ ] I need to print off the `Destruction Report` before I do the `Destroy`.  The `Report` button seems disabled until the Destroy is complete.  I should be able to view `Report` before I click `Destroy`
* [ ] We need to show a confirm box before `Destroy` is run.  So: dialog with `Are you sure xxx items have been sent for destruction?` - with `Confirm` and `Cancel` buttons.  On run `destroy` change on `Confirm`.
* [ ] In `ItemList` we should not be able to do `Loan`, `Change Location` or `Destroy` for an item that is part of a destruction job.
* [ ] In `ItemList` we should not be able to do `Destroy` for an item that is on loan to someone.
* [ ] We need to reduce the with of the `ItemList` in the `DestructionShow`. By default, just show `Reference`, `mediaType` and `Protective Marking` columns.
* [ ] For the audit log, when we add/remove an item to a destruction job, we should _also_ create audit events for the `destruction job` (we just do it for the `item` currently).

IanMayo:
Nearly there :-)

When we add an item to a `Destruction Job`, it _looks-like_ the "destruction date" is set to the current date.  But, it has not yet been destroyed.  

So, when we click on "Destroy" for the item - we set the `Destruction reference` field, but not the `destruction date`.  We only set the `Destruction Date` when the item is actually destroyed.

I _think_ this is stopping me from being able to `remove` and item from a destruction.

IanMayo:
An issue @salony75way .  The first two destructions are given the same ID:
image

IanMayo:
One small item.  After `Create` on `Destruction Job` - we stay in the `Create` form, with a new `ID` displayed.

This is _probably_ not a bug, but it is confusing.  So, after we have `Created` a new job, we should move to `DestructionShow` for the new item

IanMayo:
I have added a item, to a destruction job.  But, when I select it in `ItemList` - I am still offered `Destroy`, `Change Location` and `Loan`.  I shouldn't be (see earlier comment)

IanMayo:
There is an issue creating new destruction jobs.  The second job I create has an invalid id.  Plus, on `create` it tries to `show` item id `2`, not item id `1`, so the JS falls over here because it doesn't have a record:
image

Run-through here:
https://i.gyazo.com/dac203a8a4cd4a88683433c312ab8d9e.gif

IanMayo:
The destruction report should not have `EDIT` buttons:
image

IanMayo:
Not quite there. We don't (yet) have an `Audit History` for destruction jobs.   To support this audit history, we need to make a (slight) change to how we record the audit events - since we need audit events for resource:destruction that contain the `id` for that destruction job. 

When we add an item to a destruction, we need to create two audit entries, one for `items` and one for `destruction`.

For each one, we provide the id of the item/destruction respectively,  The activity details should include the reference for the destruction / item respectively.  Note: this is the reverse of the current implementation.  So, the destruction entry needs the reference of the item, and the item entry needs the reference for the destruction.

src/providers/dataProvider/resource-callbacks/DestructionLifeCycle.ts
```diff
@@ -0,0 +1,27 @@
+import { type CreateResult, type ResourceCallbacks } from 'ra-core'
+
+import { withCreatedByAt, type AuditFunctionType } from '../dataprovider-utils'
+import { AuditType } from '../../../utils/activity-types'
+import { R_DESTRUCTION } from '../../../constants'
+
+export default (audit: AuditFunctionType): ResourceCallbacks<any> => ({
+  resource: R_DESTRUCTION,
+  beforeCreate: async (record: CreateResult<Destruction>) => {
```
IanMayo, line 9: @salony75way  - the type of `record` for `beforeCreate` should be `CreateParams`

IanMayo: Aah, we have this wrong across a lot of RCO. Will log separate issue.

IanMayo (approved):

PR #305, opened 2023-05-18
user show permissions applied

### LINKED ISSUE #282: Permissions Phase 2 (UserShow)
Once we have the `UserShow` form in place (#281), we should cover it with permissions.

`rco-users`  should be able to:  
- view `UserList`
- click on an item to view `UserShow`
- return items from `ItemList` in  `UserShow`
- click on `Depart Organisation`
- click on `Create` from `UserList`

`power-users` should be able to:
- click on `Edit` from `UserShow` (to change roles, password, active, etc)

**Update:** we can't handle all of the above just through permissions.  `rco-user` requires write permission to `Depart` a user, since it updates the `departedDate` field. But, I don't want `rco-user` to have the `edit` button on `UserShow`.

Roles with `user:edit` permission get the `Depart` button.

We then need a way of deciding who to give the `user:edit` permission to.  I guess that if the role has `user:delete` permission then we show `Edit` in `UserShow`. Note: we still don't actually include a `User:Delete` button, since it could break referential integrity.

**Note:** I'm deliberately assigning this to @TahaKhanAbdalli - so we spread out the knowledge of how the permission model works.

TahaKhanAbdalli:
Fixes #282

### DIFF
--- a/src/resources/users/UserShow.tsx
+++ b/src/resources/users/UserShow.tsx
@@ -50,7 +50,6 @@ const DepartOrganisation = ({
   }
   return (
     <Box sx={style}>
-      <Warning />
       <Typography variant='h6'>
         Are you sure this user is departing the organisation? Doing this will
         prevent items from being loaned to the user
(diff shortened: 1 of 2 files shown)


IanMayo:
@TahaKhanAbdalli - could you include this one, too please?
https://github.com/DeepBlueCLtd/RCO/issues/306

IanMayo (approved):

Issue #95, opened 2023-04-12
Typescript refactoring


IanMayo:
I suggest we make these changes:
- [x] year of receipt to number
- [x] item_number to itemNumber
- [x] vault type from `number` to `ReferenceItem['id']`
- [x] constants for resource types to singular (see below)
- [x] remove RCOResource, switch to existing RAResource (hmm, I _think_ react-admin may not like this, in the new release that types calls)
- [x] drop ProjectCode from Batch
- [x] It is useful to know the created-at and created-by for all data.  We should modify types.d.ts so all types extend ResourceWithCreation.
- [x] once all classes extend ResourceWithCreation, further refactor DataProvider so that all resources update `createdAt`.
- [x] create interfaces/types for all reference lists. They should extend `ReferenceItem[]` or `ActiveReferenceItem[]` as appropriate.

![image](https://github.com/DeepBlueCLtd/RCO/assets/1108513/d90256a0-42f5-4560-a55f-539620ecd94c)

IanMayo:
Items done, or moved to #677

PR #314, opened 2023-05-19
lifecycle handlers refactored

### LINKED ISSUE #298: Refactor lifecycle handlers to use common code
We've created a lifecycle handler for the reference data items #297 

It would be good if we could refactor our main lifecycle handlers to extend this handler - so that the audit functionality was just in one place in the code.

Where the lifecycle handler requires extra functionality for an event, it would combine this functionality with a call to the parent.

TahaKhanAbdalli:
Fixes #298


src/providers/dataProvider/resource-callbacks/ReferenceItemLifeCycle.ts
```diff
@@ -24,13 +25,43 @@ export default (
     return record
   },
   beforeUpdate: async (record: UpdateParams<RCOResource>) => {
+    let securityRelated: boolean | undefined
+    if (checkSecurityRelated === true) {
```
IanMayo, line 29: instead of passing in a boolean for whether to test, how about passing in an optional function that does the test. Then this module does not have to know the detail of how we test different types of resource?

src/providers/dataProvider/resource-callbacks/ProjectLifeCycle.ts
```diff
@@ -1,31 +1,17 @@
...
-  type UpdateResult
-} from 'ra-core'
+  type AuditFunctionType,
+  withCreatedByAt,
+  extendLifeCycle
+} from '../dataprovider-utils'
+import { type ResourceCallbacks, type CreateResult } from 'ra-core'
 import { R_PROJECTS } from '../../../constants'
 
-export default (audit: AuditFunctionType): ResourceCallbacks<any> => ({
```
IanMayo, line 10: @Taha. Note: The only actual functionality in this module is this `createdAt` stuff.  Let's not change this PR, but in #95 I'm going to switch all types to use ResourceCreatedAt.  Then we can do `withCreatedAt` in the extendLifeCycle code.

src/providers/dataProvider/resource-callbacks/UserLifeCycle.ts
```diff
@@ -31,6 +32,8 @@ const lifeCycles = (
   }
 })
```
IanMayo, line 34: this next function is clever. well done :-D

IanMayo (approved):

PR #323, opened 2023-05-22
Item create process modified

### LINKED ISSUE #313: Help user with ItemCreate process
Users were confused with the `Save and Clone` and `Save` within `Item Create` process.

`Save` should be renamed as `Save and Create`

The new item is shown, but since its cloned, the form doesn't change.

So, for `Save and Clone` we should first popup a dialog showing:
`Item V23/2023/14 has been saved. Now showing cloned item.`

For `Save and Create` we should popup a dialog showing:
`Item V23/2023/14 has been saved. Now showing blank item.`

TahaKhanAbdalli:
Fixes #313

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "crypto-js": "^4.1.1",
     "dayjs": "^1.11.7",
     "luxon": "^3.3.0",
+    "mitt": "^3.0.0",
     "preval.macro": "^5.0.0",
     "ra-data-local-forage": "^4.8.4",
     "react": "^18.2.0",
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -56,3 +56,6 @@ export const ICON_DETAILS = Notes
 
 export const APP_VERSION = 'APP_VERSION'
 export const DATA_VERSION = 'DATA_VERSION'
+
+// mitt events
+export const SAVE_EVENT = 'save'
--- a/src/providers/dataProvider/resource-callbacks/ItemLifeCycle.ts
+++ b/src/providers/dataProvider/resource-callbacks/ItemLifeCycle.ts
@@ -13,6 +13,7 @@ import {
 import { AuditType } from '../../../utils/activity-types'
 import { R_BATCHES, R_ITEMS } from '../../../constants'
 import { type UpdateParams } from 'react-admin'
+import { emitter } from '../../../resources/items/ItemForm/ItemFormToolbar'
 
 const lifeCycles = (
   audit: AuditFunctionType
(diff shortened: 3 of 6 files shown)


IanMayo:
Have been playing with this @TahaKhanAbdalli 

When I am entering a series of items, `clone` and `save` do not get set to `false` anywhere.

I _think_ that is why I was unable to do `Save and create` after doing a few `Save and clone` entries.

IanMayo:
Hello @TahaKhanAbdalli  - I _think_ I have a repeatable issue.

On the second time I do "save and create", it will not let me submit the form:
image


But, no errors are shown.

TahaKhanAbdalli:
I was also facing the same issue on other branches also. I think it's not related to this issue in particular. I'll have a look at it.

IanMayo:
Hello @TahaKhanAbdalli - I don't really like the use of the two strings.   Well, I don't like the fact that the compiler isn't checking they're valid, and I don't like the fact that they are so general (save/clone) - since if we grow use of `mittr` elsewhere in RCO I don't want to get confused over these common names.

So, instead, could we create an enum in `ItemFormToolbar` that declares the `itemSave` and `itemClone` strings, then refer to these enums across the app?

IanMayo:
That last commit doesn't look right @TahaKhanAbdalli . It _looks-like_ we are defining the events as integers in the toolbar, but they are still strings in the item-lifecycle.

This is part of the reason why I dislike _magic strings_.

TahaKhanAbdalli:
I have no way of knowing if `save and create` button is clicked or `save and clone` button is clicked. Due to this, I have to use `save` and `clone` variables. The event is being emitted from `ItemLifeCycle` where I don't have access to which button is clicked otherwise I would have emitted different event for both of the buttons.

IanMayo:
> I have no way of knowing if `save and create` button is clicked or `save and clone` button is clicked. Due to this, I have to use `save` and `clone` variables. The event is being emitted from `ItemLifeCycle` where I don't have access to which button is clicked otherwise I would have emitted different event for both of the buttons.

Sure @TahaKhanAbdalli  - i'm happy with the strategy.  I was nervous that we were sending `save`, but listening out for `SAVE`.  But, I now see that isn't the case.

I am now just nervous about a type being introduced with `mittr` code.  We use the `save` string in two locations.  If one gets accidentally changed, the app will break without us knowing.

So I would rather you declared `const SAVE_EVENT="save"` somewhere, maybe in `constants.ts`.  Then the compiler will tell us if one of the values changes.

IanMayo (approved):

PR #325, opened 2023-05-23
#312: Dispatch process

### LINKED ISSUE #312: Dispatch process
UX for dispatch process recorded [here](https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g24693a2d2ea_0_42)

Beyond the contents of slides, we will require these changes:
- we will require a new function (prob in DataProvider) to generate the unique reference for this Dispatch (uses count of Dispatch items created this year, like we do for `Batch`.
- introduce "Dispatches" resource, include in left-hand menu
- introduce "Addresses" resource to `ReferenceData` can be edited by `rco-users` and `power-users`
- create `DispatchList` component. Shows: `reference`, `dispatchedAt`, `toName`, `remarks`, `receiptReceived`.  Filters for `live` (not dispatched), `receipt pending` (receipt not received), `overdue` (later of distpatched at and lastHastenerSent more than a month ago).
- add "New Dispatch" to the welcome page
- in `ItemShow`, if the `dispatched` field is non-null, include large `Dispatched at` label at top of page. This will be a link to the relevant `dispatch` but will display the `dispatchedDate` value in `Dispatched`.
- in the welcome page, populate the `Pending Receipt Notes` table. These are dispatches where "receipt received" is null. Show dispatch reference as `Reference`, dispatchedAt date as `Dispatched`, lastHastenerSent as `Last Hastener`.  Shade row in red if `dispatchedAt` is more than one month ago.

salony75way:
fixes #312

### DIFF
--- a/src/resources/items/ItemShow.tsx
+++ b/src/resources/items/ItemShow.tsx
@@ -104,7 +104,7 @@ export default function ItemShow(): React.ReactElement {
 }
 
 function DispatchedAt(): React.ReactElement {
-  const { dispatchJob } = useRecordContext<Item>()
+  const { dispatchedDate } = useRecordContext<Item>()
 
   const dispatchedAtSx: SxProps<Theme> = (theme: Theme) => ({
     fontSize: '30px',
@@ -112,7 +112,7 @@ function DispatchedAt(): React.ReactElement {
     '& span': { fontSize: '25px' }
   })
 
-  if (typeof dispatchJob === 'undefined') return <></>
+  if (typeof dispatchedDate === 'undefined') return <></>
 
   return (
     <TopToolbarField<Item> source='dispatchJob' component='div'>


IanMayo:
* [ ] In `DispatchCraete` could we have the `To Name` as the first field, that's the logical order for the data
* [ ] In `Dispatch`, I'm offered a list of `Destruction Jobs` in the dispatch. I _think_ it's just the label that is wrong
* [ ] In `Dispatch Show`, the `Print Hastener` button should come first.  The `Hastener` button should be renamed as `Record Hastener Sent`.

IanMayo:
Could you fix the conflicts in here please @salony75way ?

IanMayo:
* [ ] In `ItemList`, `dispatched` should be `dispatch job`.  We should also create/display `Dispatch Date` when the item does get dispatched (handle like we did for destruction).
* [ ] In `DispatchForm`, can we display the dates in two columns please, so it uses less vertical space
* [ ] After a `Dispatch` is sent, the form cannot be edited.

IanMayo:
* [ ] reduce vertical whitespace in form (see below).  There appears to be a blank line between each row. This may be preventing the whole form from being viewed on one screen
* [ ] The `Created-by` field should not be editable in `FormEdit`
* [ ] When I `Edit` an existing `Dispatch`, there is a `Delete` button. There should not be.
* [ ] In we are in `Show` for `Dispatch Job`, the `To address` should not be editable
* [ ] I have created a dispatch job, and added an item to it. When I add two more items, they do not get added to the job (see recording below)
* [ ] If an item is part of a `Dispatch job`, I should not be able to dispatch it

image


screencast

IanMayo (approved):

PR #327, opened 2023-05-23
department reference resource test written

TahaKhanAbdalli:
Supports #180


TahaKhanAbdalli:
I have no way of knowing if `save and create` button is clicked or `save and clone` button is clicked. Due to this, I have to use `save` and `clone` variables. The event is being emitted from `ItemLifeCycle` where I don't have access to which button is clicked otherwise I would have emitted different event for both of the buttons.

IanMayo (approved):

PR #329, opened 2023-05-24
285 history button

### LINKED ISSUE #285: `History` button for all resources
There is value in seeing the audit history for all kinds of items.

For `Items` we've put the history on the right-hand side of the `ItemShow` component.  But, we can't afford to do this for all resource types.

So, instead we should create a `ResourceHistory` component. This will take a resource name and id.  On each resource we'll add a `History` icon action to the `Show` page (using the `history` icon from material icons).  

It will popup an `AuditList` table showing the history for that resource item. This will have a `Print` button.

We can then remove the `AuditList` from `ItemShow`.

TahaKhanAbdalli:
Fixes #285

### DIFF
--- a/src/resources/vault-locations/VaultLocationList.tsx
+++ b/src/resources/vault-locations/VaultLocationList.tsx
@@ -11,7 +11,7 @@ import ResourceHistoryModal from '../../components/ResourceHistory'
 export default function VaultLocationList(): React.ReactElement {
   const { hasAccess } = useCanAccess()
   const [open, setOpen] = useState<boolean>()
-  const [record, setRecord] = useState<ActiveReferenceItem>()
+  const [record, setRecord] = useState<ReferenceItem>()
   const hasWriteAccess = hasAccess(constants.R_VAULT_LOCATION, { write: true })
   const filter = useMemo(
     () =>
@@ -47,7 +47,7 @@ export default function VaultLocationList(): React.ReactElement {
         />
         <FunctionField
           label='History'
-          render={(record: ActiveReferenceItem) => {
+          render={(record: ReferenceItem) => {
             return (
               <IconButton
                 onClick={(e) => {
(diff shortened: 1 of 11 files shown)


IanMayo:
(I've lost this message, so re-typing)

Could we also have the `history` button for `Users` and all of the `Reference Data` resources, please?

IanMayo:
Thanks @TahaKhanAbdalli - but could you solve these issues please?

(I know it's just the linter, and the code works fine, but I'd like to keep us at zero linter issues ;-) )
image

IanMayo (changes requested):
I've just logged in as `ian` in dev-tools I can see that the audit item has been created. But, when I click `History` for the user `ian` it is empty.  This looks like a bug.

TahaKhanAbdalli:
The `login` and `logout` actions do not have any `resource` attached to them, therefore the `history` shows empty. But if you depart user or do any actions whose `target resource` is users, then it will show in the history page with items listed.

IanMayo:
Sure. The history for `UserShow` is a special case. It shows edits to that resource.  It should also include data changes conducted by that resource (including login, loan, edit).

This will let us view/review all of the interactions for that RCO user.

IanMayo:
Hello @TahaKhanAbdalli - is it possible to show user events that _either_ record changes to the details for this user, _or_ changes performed by this user, _or_ activities that relate to this user (items loaned to them).

src/resources/vault-locations/VaultLocationList.tsx
```diff
@@ -21,10 +21,6 @@ export default function VaultLocationList(): React.ReactElement {
     [record]
   )
```
IanMayo, line 23: Hello @TahaKhanAbdalli  - I've tried calling the `setOpen` state entity directly from the TSX.

It simplifies the code, but is it ok to do that?

IanMayo (approved):

PR #330, opened 2023-05-25
assign items as loan to random users

### LINKED ISSUE #204: Generate random loans
We could do with some loaned data in the database.

Maybe something like this:
- loop through items, for each:
   - 20% chance of it being loaned (`Math.random() > 0.8`)
   - assign it to a random active user
   - use some random date in the last 3 months for `loanedDate`

Note: this can't be done until we have `loanedDate` from #255

TahaKhanAbdalli:
Fixes #204

### DIFF
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -12,7 +12,9 @@ import {
   useRedirect,
   UserMenu,
   type UserMenuProps,
-  useLogout
+  useLogout,
+  useDataProvider,
+  type DataProvider
 } from 'react-admin'
 import { SideMenus } from './SideMenus'
 import Footer from './Footer'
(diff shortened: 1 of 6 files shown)


IanMayo:
@TahaKhanAbdalli  - this may not be possible, but...

Is it possible to use the `loan` method to make the loan?  Then we'll have the `audit` history that tracks the loan.

One other thing, please.  In `ItemShow`, if the item is currently loaned out,  could we add a fieldset box saying `On loan to Ian-1` ?  This would go **Above** the `View Item` fieldset - so it's the first thing the user sees on the page.  The user name would be clickable, taking us to the resource.

IanMayo (approved):

IanMayo (approved):

PR #333, opened 2023-05-29
Double click implemented on ItemList

### LINKED ISSUE #331: Double-click ItemList to show
It's quite annoying that when I try to select an item in ItemList, if I miss the check-box it goes into `ItemShow`.

If possible, I'd prefer this behaviour, please:
- single click on row causes `select`
- double-click on row causes `open` (to ItemShow)

If we can do the above for `Items` - it will probably be worth doing it for all other lists, so the UI is consistent.

But, let's review the PR for just `ItemList` first

TahaKhanAbdalli:
Fixes #331

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -14,8 +14,7 @@ import {
   AutocompleteInput,
   type SortPayload,
   useGetList,
-  type FilterPayload,
-  useRedirect
+  type FilterPayload
 } from 'react-admin'
 import SourceField from '../../components/SourceField'
 import SourceInput from '../../components/SourceInput'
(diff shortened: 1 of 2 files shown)


src/resources/items/ItemList.tsx
```diff
@@ -221,33 +223,54 @@ export default function ItemList(props?: ItemListType): React.ReactElement {
...
+    if (selectedIds.includes(id) === false) onSelect([...selectedIds, id])
+    else onSelect(selectedIds.filter((selectedId) => selectedId !== id))
+  }
+
+  const handleDoubleClick = (id: number): void => {
+    const path = `/${constants.R_ITEMS}/${id}/show`
+    redirect(path)
+  }
+
+  const handleRowClick = useDoubleClick(handleClick, handleDoubleClick)
```
IanMayo, line 245: @TahaKhanAbdalli  - is it an option to pass in the (optional) double-click path destination to `useDoubleClick()`. That will reduce a little duplication.

TahaKhanAbdalli: I was trying to keep them `decoupled` so that `useDoubleClick` hook doesn't know what to do on `single click` or `double click`. But if you say so I can handle pass the path to `useDoubleClick` hook.

IanMayo: Yes, there isn't a clear single route through refactoring.

Could we see what it looks like with these capabilities pushed into `useDoubleClick` please?  The value in getting this implementation as "thin" as possible is that it makes it quicker/easier to apply it to other lists, too.

IanMayo: Yes, there isn't a clear single route through refactoring.

Could we see what it looks like with these capabilities pushed into `useDoubleClick` please?  The value in getting this implementation as "thin" as possible is that it makes it quicker/easier to apply it to other lists, too.

IanMayo:
Thanks @TahaKhanAbdalli 

But, now that we have removed the child elements, we don't need the custom `DataList` component any more, do we?

TahaKhanAbdalli:
We do need custom `DataList` because `useDoubleClick` is using `useListContext` which is only accessible to the descendants of `List` component

IanMayo (approved):

Issue #302, opened 2023-05-17
Bug - need Hard Refresh before permissions change


IanMayo:
I used the app as a non-logged in user, checking I can't add items to a Batch.

Then I logged in as `Jason`.  But, I still couldn't add items to a Batch.

I did reload, and I then was able to add items to a Batch.

On logout/login we must somehow refresh the browser state.

PR #345, opened 2023-06-05
current resource permission checking call

### LINKED ISSUE #293: Reduce permission checking calls
Our shiny new permissions model is working satisfactorily.

But, when I add a `console.log` to `canAccess` I see the full set of permissions are being checked/reviewed for each UI update.

This _seems_ like an issue.  I _presume_ we could be just checking permissions for resource that is being viewed.

salony75way:
fixes #293

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,8 @@
     "react/react-in-jsx-scope": "off",
     "quotes": [2, "single"],
     "comma-dangle": ["error", "never"],
-    "semi": [2, "never"]
+    "semi": [2, "never"],
+    "@typescript-eslint/strict-boolean-expressions": 0
   },
   "settings": {
     "react": {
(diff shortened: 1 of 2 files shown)


IanMayo (commented):
~~@salony75way - It _looks-like_ we don't have a TypeScript definition for `accessStates` - so the compiler won't help out if one of them gets a typo.~~

No, forget that - the compiler does pick up typos

IanMayo:
Thanks @salony75way - I think this reduces the permissions calls from the `SideMenu` - but there are lots and lots of other calls to the handler.  I _guess_ we should do some cacheing inside `useCanAccess`, instead of inside `SideMenu`.

Hmm, or maybe this is premature optimisation.  What do you think @salony75way ?

salony75way:
> Thanks @salony75way - I think this reduces the permissions calls from the `SideMenu` - but there are lots and lots of other calls to the handler. I _guess_ we should do some cacheing inside `useCanAccess`, instead of inside `SideMenu`.
> 
> Hmm, or maybe this is premature optimisation. What do you think @salony75way ?

Hi @IanMayo, We can not maintain  caching in useCanAccess because if we enable this then we need to hard refresh page after new user sign-in.

For ref #302, We removed caching in this ticket so let me know if we need to find some other way.

IanMayo (approved):

PR #346, opened 2023-06-05
vault location muster list bug fixed

### LINKED ISSUE #342: Bug - for Vault-Locations, Muster list doesn't open
In Vault-Locations we can select one or more locations, then click to view a muster list.

The must list isn't opening. No console errors are reported.

TahaKhanAbdalli:
Fixes #342

### DIFF
--- a/src/components/VaultLocationReport.tsx
+++ b/src/components/VaultLocationReport.tsx
@@ -1,7 +1,5 @@
-import { Article } from '@mui/icons-material'
 import {
   Box,
-  Button,
   Table,
   Typography,
   TableHead,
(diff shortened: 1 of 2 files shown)


IanMayo:
Hello @TahaKhanAbdalli - this is a class regression error.  Something used to work (location muster report), then it stopped working without us noticing.

I hope our automated UI testing will grow to spot issues like this.  But, just as a "test-case" - were you able to see how we broke this feature?  I/we can then look out for it in the future.

TahaKhanAbdalli:
The issue was that the `muster list modal` was also in the `bulk actions` component. I placed it inside the body of the vault location list.

IanMayo (approved):

PR #348, opened 2023-06-06
drop leading zeroes

### LINKED ISSUE #326: Drop leading zeroes in generated ids
We generate ids for `batches`, `items` and `destruction jobs`.

We provide a leading zero when necessary, to make the counter 2 digits wide.

The users do not want leading zeroes, so it's `V1/2023/4` not `V01/2023/04`.

salony75way:
fixes #326


IanMayo:
Lots of conflicts here @salony75way . You may be better off applying the `0` changes in a fresh branch.

PR #352, opened 2023-06-06
grid header font weight set to bold

### LINKED ISSUE #349: Format header bar differently
I was just looking at the `ItemList` inside `DestructionShow`.   It wasn't immediately clear that the table header was actually a header.

Here is `ItemList`:
image

I suggest we either make the header text bold, or give it a thicker horizontal separator line.   Other suggestions welcome.

Aah, the same goes for the other RCO data grids, please.

TahaKhanAbdalli:
Fixes #349

### DIFF
--- a/src/components/ReferenceDataList.tsx
+++ b/src/components/ReferenceDataList.tsx
@@ -51,12 +51,6 @@ export default function ReferenceDataList({
   return (
     <List actions={<ListActions />} resource={cName}>
       <Datagrid
-        sx={{
-          '& .RaDatagrid-headerCell': {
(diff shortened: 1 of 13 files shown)


IanMayo:
Thanks @TahaKhanAbdalli  - but is it an option to use a global theme override?
https://marmelab.com/react-admin/Theming.html#global-theme-overrides

IanMayo (changes requested):
(have asked @TahaKhanAbdalli  to try a global theme change)

IanMayo (approved):

PR #354, opened 2023-06-07
double click handling

### LINKED ISSUE #340: Refactor double-click handling
We've merged a PR that introduce double-click to open items:
https://github.com/DeepBlueCLtd/RCO/pull/333/files#diff-c0c1cdeb550e7c034692ee2b6d083e86289daa261331ca186aad99be42fd8fe3

But, we have to declare the list as a new component (in order that the hook is in the class component).  If we apply `useDoubleHooks` behaviour to the other lists in RCO, I'd like to avoid generating components.

I'd like us to investigate finding a way of using the `useDoubleClick` hook without creating a new component.

salony75way:
fixes #340

### DIFF
--- a/src/components/CustomDatagridConfigurable.tsx
+++ b/src/components/CustomDatagridConfigurable.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+import {
+  DatagridConfigurable,
+  type DatagridConfigurableProps
+} from 'react-admin'
+import useDoubleClick from '../hooks/useDoubleClick'
+import { type ResourceTypes } from '../constants'
+
+export default function CustomDatagridConfigurable(
+  props: DatagridConfigurableProps & { resource: ResourceTypes }
+): React.ReactElement {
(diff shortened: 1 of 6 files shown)


src/components/CustomDatagridConfigurable.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
```
IanMayo, line 1: one minor change, please @salony75way 

`CustomDatagridConfigurable` doesn't quite explain what's happening.

Could we call it `DblClickDatagridConfigurable`  please?

IanMayo:
@salony75way - we use `useDoubleClick` in two other components, too: `UserList` and `VaultLocationList`.   But, they use `DataGrid`, not `DatagridConfigurable`.  Supporting the two different components could introduce complexity, but I think we can just switch these two components to `DatagridConfigurable`.  It won't do any harm :-)

PR #363, opened 2023-06-09
items list destroyed/dispatched handling

### LINKED ISSUE #309: Items List - handling of destroyed/dispatched items
We're starting to introduce `Destroyed` and `Dispatched` items.

Once an item is destroyed/dispatched, it should not be shown in the `Items` table.  So, filter that table accordingly.

We will add an `All Items` table/resource to the left-hand menu - which includes destroyed or dispatched items.   This `List` will introduce filters for `Destroyed` or `Dispatched`.

salony75way:
fixes #309

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -254,7 +254,20 @@ function App(): React.ReactElement {
           key={constants.R_ITEMS}
           icon={constants.ICON_ITEM}
           name={constants.R_ITEMS}
+          options={{
+            filter: {
+              dispatchJob: undefined,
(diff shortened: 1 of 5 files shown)


IanMayo:
Hmm, I think the UX needs a little improvement.

* [ ] rename `Items` to `Live Items`
* [ ] switch to `saveAs` icon for `Live items`
* [ ] move `All Items` down the list, to just above `Reference Data`

IanMayo (approved):

PR #370, opened 2023-06-09
active flag added for vault location

### LINKED ISSUE #366: Need `active` flag for vault-location
We should add `active` flag to the vault-location type.

* [x] switch `vault-location` to being `ActiveReferenceItem`
* [x] when generating the mock data, make alternate vault
* [x] include `Active` switch in vault-location show/edit form
* [x] Only show `Active` vault-locations in `ItemCreate`.

TahaKhanAbdalli:
Fixes #366


src/resources/batches/BatchForm.tsx
```diff
@@ -41,6 +41,44 @@ const schema = yup.object({
     )
 })
```
IanMayo, line 43: Hello @TahaKhanAbdalli - can you tell me why has this text block been moved around in file, please?

TahaKhanAbdalli: This component was supposed to be used in `item form` for only showing `active vault location`. One approach was to repeat this whole block but I moved it around to `export` it so that it could be used in the item form.

IanMayo (approved):

PR #372, opened 2023-06-12
add link to item in auditlist

### LINKED ISSUE #367: All navigate to relevant item in AuditHistory
In AuditList we know the resource and data-id for the item.

When we know these, we should:
* [x] make the data-id clickable, taking us to the relevant `show` page.  
* [x] display the `name` of the item (... if that's possible. I know it's actually stored in a different table, so it may be a challenge)

salony75way:
fixes #367

### DIFF
--- a/src/resources/audit/AuditList.tsx
+++ b/src/resources/audit/AuditList.tsx
@@ -9,7 +9,8 @@ import {
   NumberInput,
   useListContext,
   AutocompleteArrayInput,
-  TextInput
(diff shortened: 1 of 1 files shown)


IanMayo (changes requested):
Thanks for this, it's working well :-)

I edited the name of a platform, but the audit list still shows the id, not the platform name

IanMayo (approved):

PR #373, opened 2023-06-12
368 high volume data

### LINKED ISSUE #368: High volume test
I'm keen to see the performance of RCO with a larger number of items.

Our eventual solution will use REST to a SQLite backend, rather than in-browser data, but we may still be able to learn lessons.

So, create a new `Load data (high volume)` entry in the top-right menu. This will create:
* [x] 60 platforms (20% inactive)
* [x] 50 media types (50% inactive)
* [x] 60 projects (90% inactive)
* [x] 200 users (40% inactive)
* [x] 100 vault locations (5% inactive)
* [x] 500 batches, each with a random (6 to 60) number of items. The batches should have dates over the last 20 years.

TahaKhanAbdalli:
Fixes #368

### DIFF
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -67,6 +67,10 @@ const MyUserMenu = (props: UserMenuProps): React.ReactElement => {
     })
   }
 
+  const handleHighVolumeLoadData = (): void => {
+    loadDefaultData(undefined, provider, true).catch(console.log)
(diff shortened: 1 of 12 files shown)


IanMayo:
We don't have the high volume of `items` in here @TahaKhanAbdalli - just the original 100 items.  I want to test it with a lot more, please.

PR #385, opened 2023-06-13
to and from address shown on dispatch receipt note and hastener report

### LINKED ISSUE #376: Include from and to address on dispatch notes
We specify an address for despatch, but it doesn't get included on the form.

We should show the name and the address on receipt notes and hastener receipt notes.

We should also show the mock sender address as shown below.

**Note:** we currently (mistakenly) show a date value in the `From` field.

image

TahaKhanAbdalli:
Fixes #376

### DIFF
--- a/src/resources/dispatch/ReportData.tsx
+++ b/src/resources/dispatch/ReportData.tsx
@@ -7,6 +7,7 @@ import * as constants from '../../constants'
 import { useParams } from 'react-router-dom'
 import { Box } from '@mui/system'
 import FlexBox from '../../components/FlexBox'
+import { FROM_ADDRESS } from '../../constants'
 
 const DispatchDetail = (): React.ReactElement => {
   return (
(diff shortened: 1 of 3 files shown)


src/resources/dispatch/ReportData.tsx
```diff
@@ -10,11 +10,33 @@ import FlexBox from '../../components/FlexBox'
...
-      <FieldWithLabel label='Date' source='createdAt' />
-      <FieldWithLabel label='To' source='toName' />
+    <Show component={'div'} actions={<></>} sx={{ marginBottom: '10px' }}>
+      <FieldWithLabel
+        label='Serial No.'
+        source='reference'
+        labelStyles={{ fontSize: '1rem' }}
+        textProps={{ variant: 'h6', sx: { fontSize: '1rem', width: '50%' } }}
+      />
+      <FieldWithLabel
```
IanMayo, line 20: No, @TahaKhanAbdalli  - the from address is a fixed address for the whole app.

See `Dept A, Building A... ` in the screenshot in the parent issue.

PR #391, opened 2023-06-14
used config data from provider

### LINKED ISSUE #388: Use `ConfigTable` in UI
We now have the `configTable` resource in the mock data.

We could do with a helper in the `DataProvider` that shares the contents of a single row of `ConfigData` (as defined in `types.d.ts`).

Then we should update the UI to use the config parameters that are defined in the `Config Data`.  This is mostly for UI labels, but also for the `From` address in `Receipt Note` and `Hastener Receipt Note`.

TahaKhanAbdalli:
Fixes #388

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -51,6 +51,7 @@ function App(): React.ReactElement {
     undefined
   )
   const [loggingPref, setLoggingPref] = useState<boolean>(false)
+  const [configData, setConfigData] = useState<ConfigData | undefined>()
   const handleGetProvider = (): any => {
     if (loggingPref !== null) {
       getDataProvider(loggingPref)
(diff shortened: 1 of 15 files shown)


IanMayo:
After a quick look, there are a few places where the `projectName` isn't displayed. I changed the value to `zzzzz` and re-created mock data. There still seemed to be lots of instances of `Project`.

I guess it's worth:
- doing a global find for `projects.`. If it's the resource name, change it to `R_PROJECTS`.  If it's not - change it to `configData.projectsName`
- do a global find for `project`.  If it's the field name, keep it.  If it's the label switch to configData.projectName.  If there isn't a label for the control - use configData.projectName

IanMayo:
(oops)

IanMayo:
Aah, there's also a linter issue:
image

IanMayo:
Aah, we missed this one, too:
image

IanMayo (approved):

PR #392, opened 2023-06-14
add resources to ref data

### LINKED ISSUE #382: Introduce `Protection` model for Batch Create and Item Create
We currently have `protective marking` for batch and item resources.

We are adopting a more complex model that comprises a number of fields.

We'll have to modify the `type` to introduce these fields for `batch` and `item` resources:
- cat-code
- cat-handling
- cat-cave

Note: we will have to introduce resources for these, with buttons on the `Reference Data` pages.  We'll also have to produce mock data for them .  Let's auto-generate 8 entries, with alternating rows marked as not active.

We will also modify the forms. The forms have these items all on one row, within a `Protection` fieldset outline:
https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g251aa763eb7_0_0

The select/radio lists will contain all of the `active` items in the relevant table. We should start off with 4 or 5 in each.  I've been assured the list won't grow any longer than these.

salony75way:
fixes #382


IanMayo:
Thanks @salony75way -
* [ ] there is also the multi-select for cat-cave, please
* [ ] Could the items be in the order their original order?  They appear to be in reverse order.
* [ ] we also need to provide `protection` for `New Batch`, please.

IanMayo:
@salony75way - in `BatchCreate` and `ItemCreate` we should only show the list of `Active` protection values.

PR #396, opened 2023-06-15
feedbacks resolved

salony75way:
Feedbacks:
- When creating a destruction the “remarks” field is mandatory, is there a reason why its mandatory? The same field in Dispatch is not mandatory. Could this field be set to non mandatory?
- Could the “report” button in destructions be renamed to Destruction Cerificatate?
- Could the “print note” button in dispatches be renamed to Print Receipt
- When setting up addresses can they automatically be set to active, rather than having to active them after the event?


IanMayo:
Thanks for this @salony75way - I'm looking forward to sharing the updated link with them.  But this seems missing:

> When setting up addresses can they automatically be set to active, rather than having to active them after the event?

Addresses are still not active by default:
image

IanMayo (approved):

PR #398, opened 2023-06-15
make start,end dates optional in items

### LINKED ISSUE #393: Refine start/end dates
For a `Batch` we only need date for start and end date. We shouldn't ask the user for the time.

For an `Item` the date is compulsory, but the time element should be optional, if possible.  If it's not optional, then please set 0000 (midnight) as the default time.

I'm not quite sure how we implement this using the `DatePicker`, but maybe we have to have separate `date` and `time` pickers, with the `time` being optional.

salony75way:
fixes #393


IanMayo:
Aah, @salony75way - I didn't describe the issue clearly enough. I've rewritten it.

IanMayo (approved):

PR #404, opened 2023-06-15
Update react admin

IanMayo:
Update react-admin, so we can use the new additional type-checking features


IanMayo:
Close. We've already updated to newest version.

PR #408, opened 2023-06-16
add history button on dispatch & destruction

### LINKED ISSUE #379: Include `Audit History` for despatch and destruction items
In the Show page for `despatch` and for `destruction` we should include a history button, to review the history for that item

salony75way:
fixes #379


IanMayo:
Just one minor suggestion, please @salony75way 

We're switching `open` from `boolean` to `string`.  How about specifying the type as `report | history`?  Then the compiler will prevent typos sneaking in.

IanMayo (changes requested):
Just one minor suggestion, please @salony75way

We're switching open from boolean to string. How about specifying the type as report | history? Then the compiler will prevent typos sneaking in.

IanMayo (approved):

PR #410, opened 2023-06-16
removed db click functionality

### LINKED ISSUE #400: Alternate strategy to open show page
We've invested effort in supporting double-click to show for selectable lists.

But, in yesterday's trial ,two users were unaware that they could drill down to get more detail.  The `double-click` is a hidden behaviour.

So, I think we need an alternate strategy.  If a DataTable is selectable, we should show an `view` icon in a column to the right of the  select column.

The icon should _probably_ be the material `preview` icon.

I _don't think_ we will then need the double-click behaviour.

salony75way:
fixes #400


IanMayo:
Can we have the icon on the far-left please?

salony75way:
> Can we have the icon on the far-left column please?

@IanMayo Seems like its not possible to add new column before checkboxes 
 
Screenshot 2023-06-16 at 6 05 39 PM

IanMayo:
Sorry - just to the right of the checkboxes.

IanMayo (changes requested):
Sorry - just to the right of the checkboxes

IanMayo (approved):

PR #420, opened 2023-06-20
create itemsCatCave table

### LINKED ISSUE #412: Need to store multiple `catCave` values for Batches and Items
Users can assign more than one `catCave` values to a batch or item.

But, I'd like to avoid a many-to-many table, if I can.

The values in `catCave` will just be short country names, like: `UK`, `GER`, `FRA`, `SPA`, `BEL`, `NOR`, `SWE`.

So, I'm thinking about strategies for doing this.  We could store a series of key values as a JSON array, and parse it in the browser :  `[2, 4]`.  At page render time we'd have to look up the label to show. That won't be hard though, since there won't be many catCave values.

Or, we could just store a JSON array of the string values: [`GER`, `BEL`].     But, I guess we'd have to do some trickery to translate the data as it goes into/out-of the multi-select control.

Any thoughts?

salony75way:
fixes #412


IanMayo:
Hello @salony75way - I've been talking with the client's DBA, who works with the legacy data.  He has looked through the existing data, and has proved that we need to support many-to-many relationships between `batches`/`items` and `catCode`/`catHandle`/`catCave` tables.

I've just documented it here: https://github.com/DeepBlueCLtd/RCO/issues/421

Sorry if this impacts on the work in this PR :-(

IanMayo:
* [ ] Protective Marking to the left of `cat handling` please (see below)
* [ ] We need to allow more space for handing and cave. So, 20% for first two columns, and 30% for second two columns, please.
* [ ] Can the multiple caveats be displayed as `lozenges` please? 

image

IanMayo:
Some linter fixes, please @salony75way :
image

IanMayo:
One issue, please @salony75way .  I can add multiple catCaves.   But, when I add a catCave, the form doesn't go dirty, so I can't save until I change some other field.

IanMayo (approved):

PR #422, opened 2023-06-20
removed localforage on loadDefaultData

### LINKED ISSUE #418: Refactor data provider
We are getting near to the time to move to a real database backend.

I want us to get the DataProvider ready for the move, including this task:
- sometimes in the DataProvider we operate directly on the `localForage`.  Avoid this, where we can. Instead make all calls/assignments through the DataProvider logic.
- look for calls to `localForage` across RCO, and replace the usage with correct use of the DataProvider API

TahaKhanAbdalli:
Fixes #418

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,6 +40,7 @@ import addresses from './resources/addresses'
 import dispatch from './resources/dispatch'
 import destruction from './resources/destruction'
 import ReferenceDataShow from './resources/reference-data/ReferenceDataShow'
+import localForage from 'localforage'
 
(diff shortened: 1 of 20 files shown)


IanMayo:
I think there's an issue @TahaKhanAbdalli 

When I delete the `indexedDB` and then `Load Data`, then app is stuck in `loading`:
screencast

TahaKhanAbdalli:
The actual issue was that `loadDefaultData` was waiting on `getDataProvider` and `getDataProvider` is waiting on `loadDefaultData` so I just shifted it from `getDataProvider` to `app.tsx`. It is not a new addition just shifting from one file to other.

IanMayo (approved):

PR #433, opened 2023-06-21
added id field on the reference data

### LINKED ISSUE #426: Need `id` values for reference data
The audit log shows foreign keys for edited items, which users can't directly interpret.

See the mediaType below:
image

We need to provide access to the `id` for reference fields. So, add an `id` column to the `list` views for the reference data. It should go on the far left.

TahaKhanAbdalli:
Fixes  #426

### DIFF
--- a/src/components/ReferenceDataList.tsx
+++ b/src/components/ReferenceDataList.tsx
@@ -6,7 +6,8 @@ import {
   type Identifier,
   List,
   TopToolbar,
-  BooleanField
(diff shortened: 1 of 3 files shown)


IanMayo:
Don't know if this is new in this PR or no, but in `Reference Data` the `platforms` table is empty. But, when I add a new batch, I see there are platforms present.

IanMayo (approved):

PR #435, opened 2023-06-22
date comparison in audit modified

### LINKED ISSUE #425: Date fields mistakenly interpreted as modified in Item Edit
I've just edited a `Live Item`, to change the media type.

But, in the audit log, RCO thinks the date fields have changed. They haven't:

image

The audit log should record date changes if they are actually changed, but not if they are un-changed.

Steps to reproduce:
- run app
- login as ian/admin
- open Live Items
- open item on second row
- click on `Edit`
- change the media type
- save the item
- re-open item on the second row
- click on view history
- note that start/end dates reported as modified

image

TahaKhanAbdalli:
Fixes #425

### DIFF
--- a/src/utils/date.ts
+++ b/src/utils/date.ts
@@ -1,5 +1,2 @@
-import { DateTime } from 'luxon'
-
-export const isSameDate = (date1: Date, date2: Date): boolean => {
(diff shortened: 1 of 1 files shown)


IanMayo:
Thanks @TahaKhanAbdalli - have you tested the fix?

IanMayo (approved):

PR #443, opened 2023-06-27
Reference field filter created for platform/project

### LINKED ISSUE #389: Allow `item` search by `platform` and `project`
Platform and Project are fields in the batch.

But, frequently users will want to find items from a particular platform or project.

We _could_ pull these fields through in the backend.  Would it be very difficult to add them to `ItemList` in the front-end?  Does react-admin support that?

@salony75way said:
> Seems like react-admin doesn't support nested references so for this type of filter we need to create a custom component.

Note: we'll support this in the future by using a VIEW to collate project and platform into the item

### Strategy

* [ ] Introduce `richItem` view that adds `platformId` and `projectId` from `Batches` table
* [ ] Connect `ItemList` to `richItem`, not `item`
* [ ] Extend front-end to introduce `Platform` and `Project` columns for `ItemList`
* [ ] Introduce `Platform` and `Project` filters for `ItemList`

Note: this issue is blocked by this one: https://github.com/DeepBlueCLtd/RCO_Legacy/issues/33

Note 2: I don't _fully_ understand how we're going to do this.  I know react-admin likes each access to a specific resource to be in the same format - without columns changing.  But, this new view will have a slightly different shape.  _maybe_ we have to introduce `constants.R_RICH_ITEM` - and `ItemList` always runs from that.  _Maybe_...

**Update:**  `richItem` view added to RCO2.sqlite

salony75way:
fixes #389

### DIFF
--- a/src/components/RefFieldFilter.tsx
+++ b/src/components/RefFieldFilter.tsx
@@ -11,6 +11,8 @@ import {
   Select,
   type SelectChangeEvent
 } from '@mui/material'
+import { useEffect, useState } from 'react'
+import { type UseRefFilter } from '../hooks/useRefFilter'
 
 interface Props<T extends RaRecord> {
   refFieldTable: string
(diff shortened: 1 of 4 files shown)


IanMayo:
Compile error @salony75way :
image

IanMayo:
Thank @salony75way - just had a play. We'll need to be able to add columns for `Platform` and `Project` too, please.

salony75way:
> Thank @salony75way - just had a play. We'll need to be able to add columns for `Platform` and `Project` too, please.

@IanMayo Do you mean we need to show  `Platform` and `Project` in items list?

IanMayo:
I'd like us to be able to let users turn on columns for `platform` or `project` in `ItemList` please.

(unless that's a real challenge)

IanMayo:
This is an interesting behaviour.  If I filter on a `platform`, the `batch` filter gets turned on:
screencast

IanMayo:
note: when I ran this locally I got a batch-id in the batch column:
image

salony75way:
> note: when I ran this locally I got a batch-id in the batch column: image

@IanMayo 
Yes because at the end its applying batch filter on items and we are using batchId as source field for project and platform filters. FYI Project and  Platform are children of batch not items.

IanMayo:
Strange behaviour when sorting on platform or project:
https://i.gyazo.com/82ec6d04b10f03f0f7a945448a40632c.gif

salony75way:
> Strange behaviour when sorting on platform or project: https://i.gyazo.com/82ec6d04b10f03f0f7a945448a40632c.gif

@IanMayo Can you please elaborate a bit more this?

IanMayo:
> > Strange behaviour when sorting on platform or project: https://i.gyazo.com/82ec6d04b10f03f0f7a945448a40632c.gif
> 
> @IanMayo Can you please elaborate a bit more this?

Sure. When I ask RCO to order on `Platform`, it actually sorts on both `Platform` and `Project`.  Are they sharing a key?

salony75way:
> > > Strange behaviour when sorting on platform or project: https://i.gyazo.com/82ec6d04b10f03f0f7a945448a40632c.gif
> > 
> > 
> > @IanMayo Can you please elaborate a bit more this?
> 
> Sure. When I ask RCO to order on `Platform`, it actually sorts on both `Platform` and `Project`. Are they sharing a key?

I think its because we are using same resource `batchId` for all of these fields

IanMayo:
We've been unable to configure `react-admin` properly to handle `project` and `platform` from the `batch` within `ItemList`. We're going to pursue an alternate strategy #451.

PR #449, opened 2023-06-29
add protectionString in item & batch

### LINKED ISSUE #447: Cache protection values
Batches and Items have many-to-many relationships with code, handle, and cave tables.

This could be expensive data to populate when showing a list of batches or items.

So, when saving a `Batch` or `Item` we should collate the set of protection values, and store them in a `protection_string` field.

Similarly, in `BatchList` or `ItemList` we should have a `Protection` column that presents this cached value.  (**note:** the `Protection` label should come from the config file)

salony75way:
fixes #447


src/utils/helper.ts
```diff
@@ -0,0 +1,10 @@
+export const transformProtectionValues = (
+  data: Record<string, any>
+): Record<string, any> => {
+  const { catCave = '', catCode = '', catHandling = '', ...rest } = data
+  const item = {
+    ...rest,
+    protectionString: `${catCave}-${catCode}-${catHandling}`
```
IanMayo, line 7: this needs the `protectiveMarking` field. They're in this order:
- catCode
- protectiveMarking
- catHandle
- catCave

Note: we need to allow for multiple values of catCode, catHandle and catCave, but just one for `protectiveMarking`.

IanMayo:
@salony75way - I _don't think_ RCO is ready for this feature yet.  We need to modify the UI to allow for multiple values of `cadeCode` and `catHandle` for `batches` and `items`.

It's only once we're supporting this more complex data that we can be sure we're implementing `protectionString` properly.  So, it's probably worth pausing this PR for now, and coming back to it.

src/utils/helper.ts
salony75way: Hi @IanMayo, We have `protectiveMarking` field in items that's why we are saving only catCave, catCode and catHandling fields in this. So here do you mean we need to create a new ref table for `protectiveMarking` and then save ref id to `protectionString` field and also remove it from items/batches?

IanMayo: Sorry @salony75way - I thought I'd answered this.

We are not formally storing this data in `protectionString`.  The formal place for all of this data is in the bridging tables (for cat_code, cat_handle and cat_cave), and the `protectiveMarking` for `batches` and `items`.

The `protectionString` is just a convenience for displaying this data in `BatchList` and `ItemList` (plus _maybe_ some reports).

So, we will never parse `protectionString`.  If we need the values we'll take them from the bridging table (or `protectiveMarking` field).  We will only use it to display `Protection` in `BatchList` and `ItemShow`.

Oh, but we will update the value on `save batch` or `save item` (_maybe_ only if the protection values change)

IanMayo:
I'm having a play, @salony75way .  In `ItemForm` I am able to add multiple code, handle and cave values to an item.   But, in dev-tools, I don't see a `protectionString` for that item.

Aah, I don't think we're catching edits to existing items - maybe we're just catching the event for `Create`.

IanMayo:
(oops)

src/resources/dispatch/ReportData.tsx
```diff
@@ -32,13 +32,6 @@ const DispatchDetail = (): React.ReactElement => {
         label='Date'
         source='createdAt'
         labelStyles={{ fontSize: '1rem' }}
```
IanMayo, line 34: hmm, @salony75way  - this next block of date values were introduced in Taha´s PR, to fix the date:
https://github.com/DeepBlueCLtd/RCO/pull/458/files#diff-2a2b593950593000491a1ba2cbda9cb814687bd0216591ef7bd623f1beccc422R35

src/resources/items/index.tsx
```diff
@@ -17,13 +17,13 @@ const ItemCreate = (): React.ReactElement => {
 const ItemEdit = (): React.ReactElement => {
   return (
```
IanMayo, line 18: @salony75way - can you teach me how this next fix works?Have we just changed the order of the attributes?

IanMayo:
Where there are multiple items, they are comma separated.  We don't need the commas, thanks.
image

src/resources/batches/BatchForm.tsx
```diff
@@ -23,12 +23,13 @@ import { Typography } from '@mui/material'
 import dayjs from 'dayjs'
 import ProtectionBlockInputs from '../../components/ProtectionBlockInputs'
 import { useConfigData } from '../../utils/useConfigData'
+import { transformProtectionValues } from '../../utils/helper'
 
 const schema = yup.object({
   yearOfReceipt: yup.string().required(),
   department: yup.number().required(),
```
IanMayo, line 30: hey @salony75way - project and platform should be nullable

IanMayo:
@salony75way - I've just spotted changes that revert new features (or fixes) in `develop`.  Could you please double-check that you aren't reverting any other new code?

IanMayo:
Hello @salony75way - I don't mind the `SonarCloud` warning.

IanMayo:
superceded by #460

PR #452, opened 2023-06-30
make project/platform optional

### LINKED ISSUE #450: Platform and Project need to be optional
I have learned that Platform and Project should be optional for batches.

* [x] modify `types.d` so that these fields are optional
* [x] modify data generation so that 30% chance of platform and 30% chance of project being null
* [x] modify BatchCreate form to reflect this
* [x] check that BatchList page works, and can still filter on project/platform
* [x] check that BatchShow page works
* [x] check that ItemList can still search on batch

salony75way:
fixes #450

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,9 @@
     "vite": "^4.2.0"
   },
   "lint-staged": {
+    "**/*.{ts,tsx,js,jsx}": [
+      "eslint --fix --ignore-path .eslintignore"
+    ],
     "**/*.{ts,tsx,js,jsx,json,md,html,css,less}": [
       "prettier --write --ignore-path .prettierignore"
     ]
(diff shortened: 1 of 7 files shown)


IanMayo:
Hello @salony75way - I've pushed a commit that gives some null platform and projects for the generated data (your code just worked in tests).

Now we have batches with missing data some console bugs are emerging when you open `BatchShow` for a batch missing platform or project.  Plus, it _looks-like_ changes aren't persisted when saving a batch after editing platform/project.  If you go to `Welcome` and back, you'll see previous version.

package.json
```diff
@@ -78,6 +78,9 @@
     "vite": "^4.2.0"
   },
   "lint-staged": {
```
IanMayo, line 80: Hello @salony75way - can you explain what this change does please?

IanMayo (approved):

PR #453, opened 2023-06-30
Inject project/platform for item

### LINKED ISSUE #451: Inject `platform` and `project` into `Item` records
For platform/project, the connection to `item` is via the `batch`.  But, users wish to be able to search items on platform/project.

@salony75way came up with a strategy to do this, but it has introduced some react-admin glitches.

We're going to take an alternate approach.  For the SQL backend, `items` is going to be a `view` that combines `item` and `batch`, pulling in `platform` and `project`.  This will make these fields available to `ItemList` (including as columns and filters).

For our mock data, we're going to:
* [x] add these optional fields to `item` in `types.d`.
* [x] populate these fields from `batch` when generating mock `item` data
* [x] populate these fields from `batch` when handling `ItemCreate`

salony75way:
fixes #451


src/resources/items/ItemForm/ItemFormToolbar.tsx
```diff
@@ -49,11 +51,18 @@ const ItemFormToolbar = (props: Props): React.ReactElement => {
...
   }, [])
 
-  const transformResource = (
+  const transformResource = async (
     data: Record<string, any>
-  ): Record<string, any> => {
+  ): Promise<Record<string, any>> => {
     const { catCave, catCode, catHandling, ...rest } = data
-    return rest
+
```
IanMayo, line 58: Hello @salony75way - how about only pulling in the fields from the `batch` if both of them are null?   If one of them is not-null then we know the backend is pulling in the extra data, and we don't need this trick.

salony75way: > Hello @salony75way - how about only pulling in the fields from the `batch` if both of them are null? If one of them is not-null then we know the backend is pulling in the extra data, and we don't need this trick.

Hi @IanMayo, Its not clear to me, Can you please elaborate a bit more?

IanMayo:
Just the comment to insert and the conflict to resolve, please @salony75way

IanMayo:
One last step, please @salony75way - we need to allow project or platform to be displayed as columns in `ItemList`, and also to be available as filters.

Oh.  We should probably display it in `ItemShow` too. Will update mockup.

PR #460, opened 2023-07-05
code refactor for protection string

### LINKED ISSUE #447: Cache protection values

salony75way:
fixes #447


IanMayo:
Thanks @salony75way - does this replace #449 ?

salony75way:
> Thanks @salony75way - does this replace #449 ?

yes

IanMayo:
@salony75way - we need to remove the commas from the combined string:
image

PR #466, opened 2023-07-05
store text value for code, handle, cave

### LINKED ISSUE #459: Have to store text value for code, handle, cave
We are using bridging tables to allow for multiple values for cat_code, cat_handle and cat_cave for batches/items.

We have tables of values for code, cat, cave.  These have an id value, a label, and an active flag.

To match the structure of the data we're porting over, the id value for these lookups is a short version of the value. So we need these changes:
- change the id from an integer to a string like "code:1", the name will be unchanged "Cat Code:1" and `active` will remain unchanged
- change the bridging tables to store the string value of id, instead of int

The rest of RCO should _just-work_.

salony75way:
fixes #459

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -59,6 +59,7 @@ interface RCOResource {
 interface ActiveReferenceItem extends RCOResource {
   // when false, the item should not be included in drop-downs
   // for `create` forms, though it should for `edit` forms
+  id: string | number
   name: string
   active: boolean
 }
--- a/src/utils/init-data.ts
+++ b/src/utils/init-data.ts
@@ -31,13 +31,25 @@ export const encryptedUsers = (isHigh?: boolean): User[] => {
   return mappedUsers
 }
 
-export const getActiveReferenceData = (
-  nameVal: string,
+type CustomIdFieldType = 'protection' | 'other'
+
+interface GetActiveRefData {
+  nameVal: string
+  alternateInactive?: boolean
+  length?: number
+  isHigh?: boolean
+  inActivePercentage?: number
+  customIdField?: CustomIdFieldType
+}
+
+export const getActiveReferenceData = ({
+  nameVal,
   alternateInactive = false,
   length = 5,
-  isHigh?: boolean,
-  inActivePercentage?: number
-): ActiveReferenceItem[] => {
+  isHigh,
+  inActivePercentage,
+  customIdField
+}: GetActiveRefData): ActiveReferenceItem[] => {
   return Array(length)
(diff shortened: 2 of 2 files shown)


IanMayo:
I see that we are now using a string as `id` for `catCave`, but in the `ItemCave` we are storing an integer id:
image

In `ItemCave` we need to store a string id for `catCave`.

salony75way:
> I see that we are now using a string as `id` for `catCave`, but in the `ItemCave` we are storing an integer id: image
> 
> In `ItemCave` we need to store a string id for `catCave`.

Yes, i will fix this

src/types.d.ts
```diff
@@ -59,6 +59,13 @@ interface RCOResource {
...
+  id: number
+  name: string
+  active: boolean
+}
+
+interface ReferenceItem extends RCOResource {
+  id: string
   name: string
   active: boolean
 }
```
IanMayo, line 71: Could we refactor this a little please?

I guess we'll have `ActiveItem` which as `name` and `active` fields.  Then we'll extend that with `ActiveItemInt` (with number id) and `ActiveItemString (with string id).

IanMayo (approved):

IanMayo:
oh, is this ready for review @salony75way ?

salony75way:
> oh, is this ready for review @salony75way ?

yes

IanMayo:
It falls over when I try to assign some cat-Code values for an item:
https://i.gyazo.com/cae09bc3fb9ad4fc4279416e01fbbbe4.mp4

It also seems to fall over when I press `Create` when adding a new Batch.  Maybe it's the code to cache the `protectionString`?

salony75way:
> It falls over when I try to assign some cat-Code values for an item: https://i.gyazo.com/cae09bc3fb9ad4fc4279416e01fbbbe4.mp4
> 
> It also seems to fall over when I press `Create` when adding a new Batch. Maybe it's the code to cache the `protectionString`?
> 
> I think we need some general testing around creating/editing batches and items.


https://github.com/DeepBlueCLtd/RCO/assets/45573884/1afaf6a3-cea3-48b6-af52-3fb57e78cde0
@IanMayo Its working on end, Have you tested with latest code?
Can you please test this once more and let me know if you are still getting this issue?

IanMayo:
It just failed when creating a new batch in the Heroku deployment for this branch:

failing create batch item

IanMayo:
Thanks for the update @salony75way - we have an issue related to the refactoring of organisations and departments:
image

PR #467, opened 2023-07-05
user should stay on item create after creating item

### LINKED ISSUE #461: When adding an item to a batch, stay in `Add item` view
Note: this is when we're adding an item to a batch, not editing an item.

Users will enter large numbers of items to a batch.  So, once an item is added, and the user clicks on `Save and clone` or `Save and create` the user should stay in the `ItemCreate` view.

TahaKhanAbdalli:
fixes #461

### DIFF
--- a/src/components/ProtectionRefInput.tsx
+++ b/src/components/ProtectionRefInput.tsx
@@ -129,7 +129,7 @@ export default function ProtectionRefInput<
       if (record?.id) {
         updateRecord(record.id as number, selectedData)
       } else if (id) {
-        if (resource !== constants.R_ITEMS) createRecord(id, selectedData)
+        createRecord(id, selectedData)
       }
       if (resource === constants.R_BATCHES) {
         const path = `/${constants.R_BATCHES}/${id ?? record?.id}/show`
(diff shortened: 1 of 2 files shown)


src/components/ProtectionRefInput.tsx
```diff
@@ -129,7 +129,7 @@ export default function ProtectionRefInput<
       if (record?.id) {
         updateRecord(record.id as number, selectedData)
       } else if (id) {
-        createRecord(id, selectedData)
+        if (resource !== constants.R_ITEMS) createRecord(id, selectedData)
       }
```
IanMayo, line 133: Hello @TahaKhanAbdalli - just to help my understanding... Can you teach me how this fixes the issue?

TahaKhanAbdalli: My mistake, I was to stop redirecting only in the `useRefTable` hook if `resource = R_ITEMS`

TahaKhanAbdalli:
Codacy issue is because of `linter` it automatically adds `await` when commiting.

IanMayo (approved):

PR #469, opened 2023-07-06
fixed repeated refTable entries

### LINKED ISSUE #463: Test bridging table entries created for `Save and clone`
When adding items to a batch, we allow the user to save/create and create a clone of the current item.

Since creating the `clone` behaviour, we now use `bridging tables` to allow multiple cat-code, cat-handle and cat-cave values to be specified.

Trial creating an item with multiple values for these, then create/clone and add another item with slightly different protection values.

Check that new bridging table entries are created for the two different items.

Note: I've just tested this. When we do `Create/Clone` the `code, handle, cave` fields for the previous item get duplicated:
image

TahaKhanAbdalli:
fixes #463

### DIFF
--- a/src/resources/items/ItemForm/CoreForm.tsx
+++ b/src/resources/items/ItemForm/CoreForm.tsx
@@ -21,11 +21,15 @@ interface Props {
   batchId?: number
   disabled?: boolean
   itemId?: Item['id']
+  setItemId: React.Dispatch<React.SetStateAction<number | undefined>>
 }
 
 const CoreForm = (props: Props): React.ReactElement => {
-  const { batchId, disabled, itemId } = props
+  const { batchId, disabled, itemId, setItemId } = props
   const formContext = useFormContext()
+  const {
+    formState: { isSubmitted, isSubmitting }
+  } = formContext
   const [mediaTypes, setMediaTypes] = useState<any[]>([])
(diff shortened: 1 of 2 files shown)


src/resources/items/ItemForm/CoreForm.tsx
```diff
@@ -37,6 +41,12 @@ const CoreForm = (props: Props): React.ReactElement => {
     setMediaTypes(data)
   }, [data])
 
+  useEffect(() => {
+    if (isSubmitted) {
+      setItemId(undefined)
+    }
+  }, [isSubmitted, isSubmitting])
```
IanMayo, line 48: hmm, I'm at my limits of `react` here, but...

If the handler doesn't do anything with `isSubmitting`, then is there any point in listening out for it?

TahaKhanAbdalli: If it's not there it does not reset the id for the `save / create`.

IanMayo: aah, ok.  So, the state of `isSubmitted` may not have changed, but we still need to clear the item. Got it!

IanMayo:
hello @TahaKhanAbdalli - just to help my understanding, can you tell me what was causing the double-submission, and how this PR fixes it please?

IanMayo (approved):

PR #474, opened 2023-07-12
integrated dataprovider to work with sqlite using soul

### LINKED ISSUE #471: Transition to SQLite backend
Our in-memory database is just for development.

We're eventually going to connect to `SQLite` via the `soul-cli` REST wrapper.

Here is the repo: https://github.com/DeepBlueCLtd/RCO-Soul

It currently has a database with the wrong schema, I hope to provide a valid database tomorrow.

The readme includes instructions on how to run the server.   I propose we take this strategy:
- clone rco-soul
- run install for rco-soul
- run dev:soul
- modify RCO to integrate with the soul-server

Note: there is a `react-admin` app in the `src` folder. There is a working `dataProvider.ts` in the `src/common` folder.  This includes functionality to implement the correct REST API calls.

I _think_ the RCO data provider will have to pass its calls through this `dataProvider`.   We'll copy the `dataProvider` into `RCO` repo.

Bit's missing from `rco2.sqlite`:
* [x] config data
* [x] rename `catHandling` to `catHandle`
* [x] `IDs` need to be `integer` for references to work
* [x] dispatch table doesn't have `toAddress` field
* [x] `IDs` should be auto-incremented
* [x] Project table doesn't include `start and end dates`
* [x] Change `rco_power_user` in user table to `rco-power-user`
* [x] add at least `5 project` entries in `batch table` to test filters
* [x] `Vault` table is not there
* [x] `User` table doesn't have `salt` field which is required for `password security`
* [x] The schema of `vault` is not valid
* [x] `dataId` is not there in `audit` table
* [ ] ~~Change `batch` in item table to `batchId`~~ The front end has been changed to `batch`.
* [x] configData has `snake_case` fields (e.g. cat_code, cat_handle, cat_cave)
* [x] there is `item_number` (snake_case) in item type on frontend, need to convert to `itemNumber`
* [ ] `Start` and `End` dates in the `database` are not according to `constraints` (start must be before end) in `project` table. Remarks are required in `edit project` but not in `create project`. 
* [x] `Remarks` is not required on `dispatch form` on frontend but database has `Not Null constraint` on remarks
* [x] `Dispatch` table doesn't have `reportPrintedAt`
* [x] `Destruction` table doesn't have `reportPrintedAt`
* [ ] need to fix `BooleanFilter`

**Need to update the `dataProvider` from RCO-Soul for accomodating filters (e.g. `_eq`, `_neq`)**

**TODO:**

- [x] Filters for handling _eq, _neq etc.
- [x] Destruction remarks on `frontend` is marked as optional whereas database has `Not NULL` constraint on remarks
- [x] Item `remarks` and `musterRemarks` on `frontend` are marked as optional whereas database has `Not NULL` constraint on remarks and musterRemarks

TahaKhanAbdalli:
Fixes #471 

To RUN `rco-soul` backend:

- clone [rco-soul](https://github.com/DeepBlueCLtd/RCO-Soul) repo.
- copy latest `RCO2.sqlite` to `rco-soul/db`
- overwrite `rco-soul` script `"dev:soul": "dotenv -- cross-var soul -d ./db/RCO2.sqlite -p $PORT"` for linux
- run rco-soul: `npm run dev:soul -p 4000`


IanMayo:
@TahaKhanAbdalli - we have a build failure for this PR.  Could you investigate it please?

src/providers/dataProvider/index.ts
```diff
@@ -135,8 +135,17 @@ export const dataProvider = (apiUrl: string): DataProvider => ({
   },
```
IanMayo, line 136: ok, for next week.  Let's define these as a const array somewhere in this file. Call it "Tables_with_string_PK".

We should probably use this code for all `Many` verbs.

src/utils/init-data.ts
```diff
@@ -1,17 +1,7 @@
 import users from '../providers/dataProvider/defaults/users'
 import { generateSalt, encryptData } from './encryption'
```
IanMayo, line 2: @TahaKhanAbdalli - I don't want to remove the concept of mock data. There are times when it could be useful to have in-browser dataset.

I just want us to have a flag (in `app.tsx`?) that indicates whether to use `localForge` or `REST` backend.

src/providers/dataProvider/index.ts
TahaKhanAbdalli: We don't need that anymore as the issue in `rco-soul` for string `ids` has been resolved

IanMayo:
Afraid I don't know the answer to the conflict @TahaKhanAbdalli - I'll have to leave it to you (it's only minor, I'm sure)

package.json
```diff
@@ -9,7 +9,7 @@
   "homepage": ".",
   "scripts": {
     "dev": "vite --open",
-    "dev:mock": "MOCK=1 vite --open",
+    "dev:mock": "MOCK=0 vite --open",
```
IanMayo, line 12: @TahaKhanAbdalli - it would be more logical for `MOCK` to take value of `1`.  `0` tends to equal `False`.

Do we need to assign value to `MOCK`?  Can't we just have `MOCK` in command-line?  When we decide the backend, we know to use `REST` if mock is falsy.

TahaKhanAbdalli: > @TahaKhanAbdalli - it would be more logical for `MOCK` to take value of `1`. `0` tends to equal `False`.
> 
> Do we need to assign value to `MOCK`? Can't we just have `MOCK` in command-line? When we decide the backend, we know to use `REST` if mock is falsy.

Now only need to set `MOCK` from command line no matter the value and run `yarn run dev`, this will run `MOCK` and to run `REST`, we don't need to set `MOCK` from command line.

src/resources/items/ItemList.tsx
```diff
@@ -145,14 +145,14 @@ const getFilters = (resource?: string): React.ReactElement[] => {
   if (resource === constants.R_ALL_ITEMS) {
     filters.push(
       <BooleanFilter<Item>
```
IanMayo, line 147: @TahaKhanAbdalli - we need the previous objects for these.

When something is added to a scheduled dispatch or destruction, the `destruction` and `dispatch` ids get set. But, we only know it has been `dispatched` or `destroyed` when the `...Date` has been set.

Was this change introduced to resolve some change?

IanMayo (approved):
Some outstanding problems, but they will be resolved as separate issues.

PR #476, opened 2023-07-12
allowed batch and item start and dates to be null

### LINKED ISSUE #472: `Item` start and end dates can be `null` (remove start/end for `Batch`)
I've seen that a lot of legacy data has null for start and end dates.

We won't allow this in new data, but we need to make sure the legacy data remains searchable.

Modify `types.d` so that these dates can be `text` or `null`.

Change data generation so that, hmm, 5% of items  have `null` for both start and end dates.

**Update:** based upon further data analysis, since Batch fields are optional, I'm pushing to remove them, since they don't inform any business processes.  Clients have agreed to removal of `start` and `end` for `Batch` elements.

TahaKhanAbdalli:
fixes #472


IanMayo:
Thanks @TahaKhanAbdalli  - this all looks ok.  Are there any runtime problems with BatchList or ItemList?  Do the time filters still work ok?

TahaKhanAbdalli:
> Thanks @TahaKhanAbdalli - this all looks ok. Are there any runtime problems with BatchList or ItemList? Do the time filters still work ok?

No they are not working and I just checked on the main branch, they were not working already.

IanMayo:
Based on further analysis, I am aiming to remove start and end dates for batches - it appears no-one requires or uses this data.
 I'll close this for now, to keep the list of PRs clean - and we'll come back to it if it does turn out that we need to store this data.

PR #477, opened 2023-07-12
introduce vault section

### LINKED ISSUE #473: Introduce concept of `Vault`
In the legacy data there were two vaults where items were stored.

Vault is an attribute of `batch`.

So, for this legacy data, the first item of batch 15 in 2023 in the main vault will have id: `VAULT/V/15/2023/1`

Introduce new table, containing two vaults (`Vault` and `Vault Legacy`).  Only the first one is active.

* [x] add new table in `types.d` for `vault`.  It will have a single text field used as id, and an active flag
* [x] add new field in `batch` for `vault`
* [x] generate two vaults in data generation
* [x] randomly assign batches to either vault in data generation
* [x] in `BatchCreate` include drop-down for vault.  If only one active, choose it by default (like we do for department/organisation)
* [ ] Include `Vault` in `BatchShow`  (mockup [here](https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g1e4271274fd_0_0))
* [x] include `Vault` in BatchList columns and filters (but, don't show it by default)

Updated schema [here](https://docs.google.com/drawings/d/1SYIb0PZxcabvhXdNMAEHlbdWaoam2ipB7x5ZUudVVGc/edit?usp=sharing)

salony75way:
fixes #473


IanMayo:
In the `BatchCreate` form, we shouldn't have the leading hash for the vault items:
image

Also, in `BatchCreate` we should only show active vaults.  Only one is active, so it should be pre-selected.

We need the `Vaults` table to be added to the `Reference Data` page

We need `Vault` to be added to columns and filter for `BatchList` page

IanMayo (approved):

PR #515, opened 2023-07-18
vault converted to referenceItem instead of ActiveReferenceItem

TahaKhanAbdalli:


IanMayo:
Hello @TahaKhanAbdalli - I've done some refactoring, and pulled the changes into here.

I've reinstated the `Name` field for `Vault` - since without a name lots of things were breaking.

Now there are _only_ these things breaking:
- in BatchList we don't show items in the `Vault` column
- when we edit an existing batch, it doesn't show the existing vault value. Plus, it doesn't validate, since the form is expecting an integer

It would be useful if you're able to progress these tomorrow, else I will

TahaKhanAbdalli:
> Hello @TahaKhanAbdalli - I've done some refactoring, and pulled the changes into here.
> 
> I've reinstated the `Name` field for `Vault` - since without a name lots of things were breaking.
> 
> Now there are _only_ these things breaking:
> 
> * in BatchList we don't show items in the `Vault` column
> * when we edit an existing batch, it doesn't show the existing vault value. Plus, it doesn't validate, since the form is expecting an integer
> 
> It would be useful if you're able to progress these tomorrow, else I will

I'm working on it.

IanMayo (approved):

PR #518, opened 2023-07-19
double check receipt note

### LINKED ISSUE #492: Double-check receipt note printed before Dispatch
User was able to finish dispatch without creating a receipt note.

Once the dispatch was complete, they couldn't print the receipt note.

The same would be an issue for destruciton too.

So, capture date of opening receipt note/ destruction certificate.  Record the new date field in `types.d`. Call it `reportPrinted`.

Only enable confirm button once sheet has been printed off.

salony75way:
fixes #492


src/types.d.ts
```diff
@@ -243,6 +243,7 @@ interface Destruction {
   finalisedAt?: string
   finalisedBy?: User['id']
```
IanMayo, line 244: @salony75way for the two new `reportPrinted` fields, could you rename them to `reportPrintedAt` - to match other date fields please?

IanMayo (approved):

PR #530, opened 2023-07-19
dispatch audits issue fixed

### LINKED ISSUE #513: Tidying to dispatch process
When items are added to a dispatch, we create an Audit Log entry for the item:
> Add item to dispatch V/2/2023

When the dispatch is completed, we _should_ create an audit log for the `item` with an `Activity` of `Sent` saying:
> Item in completed dispatch V/2/2023

But, it currently says:
> add item to dispatch

(note lower-case `add`)

Also, when we complete a dispatch, the audit log for the dispatch tries to show that the `Date Dispatched` has been edited.  But, it omits the field name.

It would actually be better to include an Audit entry for `Dispatch Sent` - as recorded in the events here:
image

salony75way:
fixes #513

### DIFF
--- a/src/providers/dataProvider/dataprovider-utils.ts
+++ b/src/providers/dataProvider/dataprovider-utils.ts
@@ -68,10 +68,10 @@ export const auditForUpdatedChanges = async (
   const difference = getDifference(record.data, record.previousData)
   const keys = Object.keys(difference)
   if (keys.includes('dispatchedAt') && !difference?.dispatchedAt) {
-    difference.dispatchedAt = null
+    difference.dispatchedAt = 'unset'
   }
   if (keys.includes('reportPrintedAt') && !difference?.reportPrintedAt) {
-    difference.reportPrintedAt = null
+    difference.reportPrintedAt = 'unset'
   }
   const dataId =
     record.previousData.id !== undefined ? record.previousData.id : null
--- a/src/providers/dataProvider/dataprovider-utils.ts
+++ b/src/providers/dataProvider/dataprovider-utils.ts
@@ -67,12 +67,17 @@ export const auditForUpdatedChanges = async (
 ): Promise<UpdateParams<RCOResource>> => {
   const difference = getDifference(record.data, record.previousData)
   const keys = Object.keys(difference)
-  if (keys.includes('dispatchedAt') && !difference?.dispatchedAt) {
-    difference.dispatchedAt = 'unset'
-  }
-  if (keys.includes('reportPrintedAt') && !difference?.reportPrintedAt) {
-    difference.reportPrintedAt = 'unset'
-  }
+  const testKeys: string[] = [
+    'dispatchedAt',
+    'reportPrintedAt',
+    'lastHastenerSent',
+    'receiptReceived'
+  ]
+  testKeys.forEach((key) => {
+    if (keys.includes(key) && !difference?.[key]) {
+      difference[key] = 'unset'
+    }
+  })
   const dataId =
     record.previousData.id !== undefined ? record.previousData.id : null
   await audit({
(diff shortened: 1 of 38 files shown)


IanMayo:
Thanks @salony75way .  But, there is still an issue. When I look at the history of a completed dispatch, it fails to show the changes to the printedAt and dispatchedAt fields. Note (there is probably one of the date edits not recorded):
image

They will _probably_ have been null values previously, but it is ok to show `Previous values: {dispatchedAt: null}`

src/providers/dataProvider/dataprovider-utils.ts
```diff
@@ -66,6 +66,12 @@ export const auditForUpdatedChanges = async (
   subject?: User['id']
 ): Promise<UpdateParams<RCOResource>> => {
   const difference = getDifference(record.data, record.previousData)
+  const keys = Object.keys(difference)
```
IanMayo, line 69: @salony75way - I'm nervous that a subsequent change to dispatchedAt or reportPrintedAt could report that the previous values were null, when they weren't.  
The current UI prevents this happening, but we don't know what will happen with the UI in the future.  So, I'd rather we check that there was no value previously present, please.

src/resources/dispatch/DispatchShow.tsx
```diff
@@ -192,19 +192,23 @@ export default function DispatchShow(): React.ReactElement {
...
     await audit(audiData)
-    await audit({
-      ...audiData,
-      resource: constants.R_DISPATCH
-    })
   }
 
   const dispatch = async (data: UpdateParams): Promise<void> => {
+    const audiData = {
+      type: AuditType.EDIT,
```
IanMayo, line 205: could we introduce an AuditType.SENT please, and use that value in as the `type` please?

src/providers/dataProvider/dataprovider-utils.ts
salony75way: Done!

src/providers/dataProvider/dataprovider-utils.ts
```diff
@@ -66,6 +66,13 @@ export const auditForUpdatedChanges = async (
...
+  if (keys.includes('dispatchedAt') && !difference?.dispatchedAt) {
```
IanMayo, line 70: Hello @salony75way - to make the report more human-legible, is it possible to specify the value as `"unset"` instead of `null`?

salony75way: @IanMayo Should we only need to add "unset" to audit log or we need to set "unset" to db too? JFY we can't set "unset" to db as this field's data type is date. Please confirm this.

IanMayo: oh, I thought we were just generating this field to go into the remarks text.

IanMayo:
Sorry @salony75way - we also need the "date test" for `lastHastenerSent` and `receiptReceived`, please.

Oh and `Hastener Sent` event should be of `activityType.Sent`, not `activityType.Edit`, please.

src/providers/dataProvider/dataprovider-utils.ts
salony75way: Ok

salony75way:
> Sorry @salony75way - we also need the "date test" for `lastHastenerSent` and `receiptReceived`, please.
> 
> Oh and `Hastener Sent` event should be of `activityType.Sent`, not `activityType.Edit`, please.

@IanMayo "date test" is not clear to me, Can you please elaborate a bit more, What we need to do here?

IanMayo:
> @IanMayo "date test" is not clear to me, Can you please elaborate a bit more, What we need to do here?

Sorry @salony75way .  We have introduced a test for two date fields being modified, and that they were previously null.  I'd like us to also do that handling for lastHastenerSent and receiptReceived, please. This is to avoid users not understanding what a previous value of `null` would be.  I do expect them to understand that the previous value was `unset`.

IanMayo:
Thanks @salony75way - I'm trying to review on [Heroku](https://rco-review-pr-530.herokuapp.com/), but it's crashing.
image

IanMayo:
Note: I'm getting the same crash when I run locally.

PR #538, opened 2023-07-20
introduce editRemarks on edit item

### LINKED ISSUE #533: Introduce `Remarks` when saving an edited `Item`
We create an audit log entry when an item is edited.

We should invite users to explain why they have made the change. 

So, on clicking `Save`, popup a form for user to explain the edit.

* Title: `Batch Item editing remarks`
* Text box with `Remarks here` text, waiting for text to be entered. Should probably just need a 4-line text area.
* Cancel button closes popup, leaves form in dirty state (with `Save` enabled)
* Save button closes popup, and performs `Save`

When the save is performed the audit log will contain both the previous values of items edited, plus the remarks text.

salony75way:
fixes #533

### DIFF
--- a/db/Scripts/full.sql
+++ b/db/Scripts/full.sql
@@ -301,8 +301,8 @@ CREATE TABLE if not exists  configData (
   projectsName  TEXT NOT NULL,
   fromAddress  TEXT NOT NULL,
   protectionName  TEXT NOT NULL,
-  cat_code  TEXT NOT NULL,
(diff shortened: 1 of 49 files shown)


IanMayo:
Working well @salony75way 

One minor request: when the remarks dialog opens, could the cursor be inside the text box please?

IanMayo:
I'm happy with this, but can you fix the conflicts please @salony75way ?

PR #540, opened 2023-07-20
user to be allowed only single role instead of multiple roles

TahaKhanAbdalli:


IanMayo:
Something to check, please @TahaKhanAbdalli   Only power users are able to add or edit platforms.  But, while logged in as `jason` I was able to edit a platform name.

I don't know if this is a legacy bug or a new one, but could you investigate, please?

IanMayo:
I'm playing with this.  These parts of the implementation don't match the original [spec](https://github.com/DeepBlueCLtd/RCO/issues/173):
* It looks like rco-user should be able to add a new project, but I couldn't.
* rco-user should be able to create/edit dispatches and destructions

src/providers/authProvider/permissions.ts
```diff
@@ -26,51 +28,9 @@ const permissions: Record<UserRoles, ResourcePermissions> = {
 
 export default permissions
```
IanMayo, line 30: @TahaKhanAbdalli  - we've removed a large block of code here.

Can you give me a summary of why we've been able to do that?

TahaKhanAbdalli: This was being used for getting permissions of the multiple roles assigned to the user. But now as we are using only one role therefore, we don't need to merge permissions.

TahaKhanAbdalli: And as the previous code was only sending permissions[role] if only one role was provided therefore, indicating that the removed code was being used for multiple roles 
image

IanMayo (approved):

PR #556, opened 2023-07-24
541 item create changes

### LINKED ISSUE #541: Item Create changes
1.  button names
Change `Create/Clone` to `Save/Clone`
Change `Create/New` to `Save/New`

2. Start/finish dates
Frequently the items are consecutive recordings (like in a CCTV), possibly one per day.  So, a lot of the time, the start of the next item is almost exactly the same as the end time of the previous now.  If we provide that as initial text, frequently it will almost hte right value

So, on `Create/Clone`, set the `Start Date-time` in the new item to be the `Finish Date-time` of the current item

Mockup [updated](https://docs.google.com/presentation/d/1eCTxSe2yML3xb3y5c9uS8zm3GdotJymJIZ-6NrfrpZg/edit#slide=id.g221ed9988a1_1_133)

salony75way:
fixes #541


IanMayo:
Thanks @salony75way - but I'm reluctant to update the branch if it is going to pull in the problem what we have in `main`.  Could you handle updating this issue from `main`?

IanMayo (approved):

PR #560, opened 2023-07-25
543 composite location column for items

### LINKED ISSUE #543: Composite location column for items
To know the status of an item, users currently have to switch on `Loaned to`, `Destruction` and `Dispatched` columns.

In their current app they have one column that shows all of these.  So, create a `Location` column. This will display this content, depending on status:
- loanedTo is non-null - show the name of the person it's loaned to
- destructionDate is non-null - show `DESTROYED`
- dispatchDate is non-null - show `SENT`
- otherwise - show (human-readable) value of `Vault Location`

salony75way:
fixes #543


IanMayo (approved):
Nice piece of work, thanks @salony75way

PR #563, opened 2023-07-25
544 item created notification larger

### LINKED ISSUE #544: For `Item Created` notification should be larger, and require Click to close
When an item is created, the user has to write the new item reference onto a label on the item.

It's really important that they can easily get to the reference.

Currently, after an item is created, the new id is shown in a notification that disappears after a few seconds.  Make it much larger (so the user can confidently read the new reference), and make it stay open until the user acknowledges it via a click.

If we can't do the above via the current notification mechanism, let's create a new component, with an API that matches `useNotify`

salony75way:
fixes #544


IanMayo (changes requested):
Works great. Just the `conflict` to resolve, please @salony75way

IanMayo (approved):

Issue #565, opened 2023-07-25
Bug - can't view item


IanMayo:
I'm in the main branch of RCO, here: https://deepbluecltd.github.io/RCO/#/item

In the `Live Items` table, when I click on the `View` button for an item, it doesn't open the `Show` page for the item, nothing happens.

This is an issue in `Live Items`, `All Items` and the Item List under `Batch Show`.

Have tried it in incognito window, and it's still an issue there.

PR #568, opened 2023-07-26
audit for fk changes

### LINKED ISSUE #520: Audit History not kept for FK changes
When a user edits an `item` we capture the previous value(s) in the audit log.

But, when an entry in a bridging table is edited `ItemCode`, `ItemCave`, `ItemHandle`, then this edit is not captured.

We need to capture these changes for the `Item` in the `Audit Log`.  

Note: for changes to the above fields, we need to set the `Security Related` flag to true.

salony75way:
fixes #520


IanMayo:
Hello @salony75way - I've just changed the `catCode` and `catHandle` for a Batch, but only got this recorded:
image


I can't edit an item because I can't get an item to open, because of #565

IanMayo:
Had a play @salony75way  :-D

I changed the `catCode` for an item.  In the audit log I see two entries, there should really just be one (see below).

I'm not sure what the change in the code has to be, sorry.
image

IanMayo (approved):

PR #573, opened 2023-07-26
introduce media category

### LINKED ISSUE #562: Introduce Media Category
There are going to be multiple lists of MediaType, to keep the length short enough to be scanned at once by a user.

For this, we need to introduce a `Media Category` table.  It will contain two values initially:
- 1: Paper
- 2: Magnetic

The MediaType table will gain a non-null mediaCategory column that will take one of the above ID values.

ItemCreate, ItemShow, ItemList will gain a `Media Category` column (and filter for Item List). On the forms, it will go before the `Media Type` drop-down.  The MediaType drop-down will be filtered according to the selected MediaCategory.  Note: we will not be storing `mediaCategory` - just `media-type`.  For item edits, we'll need to `reverse-engineer` the media category value.

Our mock data generation will be modified to randomly select a Media Category, then select a random Media Type for that category.

In `Save/Clone` ensure we are duplicating the value of `MediaCategory`

TahaKhanAbdalli:
fixes #562

### DIFF
--- a/src/resources/items/ItemForm/CoreForm.tsx
+++ b/src/resources/items/ItemForm/CoreForm.tsx
@@ -3,17 +3,19 @@ import {
   DateTimeInput,
   TextField,
   TextInput,
-  useGetList
+  useGetList,
+  useRecordContext
 } from 'react-admin'
+import { TextField as MuiTextField , Autocomplete, Card, CardContent, Typography } from '@mui/material'
 import FlexBox from '../../../components/FlexBox'
 import { useFormContext } from 'react-hook-form'
 import { useEffect, useState } from 'react'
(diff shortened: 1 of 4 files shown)


IanMayo:
* [ ] Could we have the two controls on one line, please?
* [ ] When I select a `media-category`, no `media-types` are available. I suspect we have to grow `generate data` to include media-category (note, it has to be the correct category for that type).
* [ ] We need `Media Category` table in `Reference Data`
* [ ] We need `media category` field in `Media Type` table for `Reference Data`
* [ ] In `ItemList`, I can't remove a `Media Type` filter
* [ ] We need `Media Category` in `Item Show`.

IanMayo:
Close. Media Category no longer required.

PR #576, opened 2023-07-27
include hasteners in receipt notes

### LINKED ISSUE #510: Include hasteners in `Pending Receipt Notes`
On the welcome page we show a list of pending receipt notes:
image

These should only be items that have a `dispatchedAt` date, but not a `receiptReceived` date (I think we're currently ignoring the `dispatchedAt` date test - it looks like we're showing all items.

Also, when we have been waiting for 4 weeks or more for a receipt note (or since the last hastener was printed), then show the item in red.

General process shown below:
image

salony75way:
fixes #510

### DIFF
--- a/src/components/PendingReceiptNotes.tsx
+++ b/src/components/PendingReceiptNotes.tsx
@@ -1,0 +1,48 @@
+import { DateTime } from 'luxon'
+import Recent from './Recent'
+import * as constants from '../constants'
+import { DateField, type Identifier, useGetList, useTheme } from 'react-admin'
+import HastenerSentField from './HastenerSentField'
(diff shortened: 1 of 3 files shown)


IanMayo:
Thanks @salony75way 

The text of the overdue batch is difficult to read. Can we make the red a lighter shade please?
image

Maybe this color:
image

IanMayo:
One change in the business logic, please @salony75way 

We highlight the dispatch as overdue if the `last hastener sent` is more than 4 weeks (if there is one set), else it's the dispatched date.

IanMayo (approved):

PR #577, opened 2023-07-27
muster list report format

### LINKED ISSUE #485: user muster list - report format
When user pressed <crtl-p> to print, it offered a blank page, with a grey stripe:
image

User then used `RCO` <print> button to print. The document came out landscape, and the font was too large (see below).

* [ ] Verify we're forcing portrait page-mode, if we can.
* [ ] Shrink font size to about 12px
* [ ] Remove the block of user-name/permission from the report (we have the user name at the top)
* [ ] Correct the count of items 100% mustered (says 22 in this report, should be 4)

image

salony75way:
fixes #485

### DIFF
--- a/src/components/Printable/index.tsx
+++ b/src/components/Printable/index.tsx
@@ -12,6 +12,15 @@ const style = {
   '@media print': {
     '@page': {
       size: 'portrait'
+    },
+    '& .MuiBox-root *': {
+      fontSize: '12px'
+    },
+    '& .RaDatagrid-rowCell, .RaDatagrid-headerCell': {
+      padding: '0 !important;'
+    },
+    '& .RaDatagrid-rowCell *, .RaDatagrid-headerCell *': {
+      fontSize: '12px !important'
     }
   }
 }
--- a/src/components/Printable/index.tsx
+++ b/src/components/Printable/index.tsx
@@ -8,7 +8,12 @@ interface Props {
 }
 
 const style = {
-  background: '#fff'
+  background: '#fff',
+  '@media print': {
+    '@page': {
+      size: 'portrait'
+    }
+  }
 }
 
 const buttonSx = {
--- a/src/components/Printable/index.tsx
+++ b/src/components/Printable/index.tsx
@@ -13,7 +13,14 @@ const style = {
     '@page': {
       size: 'portrait'
     },
-    '& .MuiBox-root *': {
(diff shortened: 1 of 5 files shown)


IanMayo:
Are you able to reduce the font-size a little @salony75way ?

salony75way:
> Are you able to reduce the font-size a little @salony75way ?

@IanMayo Yes i did for signature text, Do i need to set for all texts (list and heading..)?

IanMayo:
yes, smaller for all text please. it's cramped when printed portrait.

IanMayo:
Thanks @salony75way - the second signature block has slipped down the page.  Could you align them again please?
image

image

Also, there is a wide horizontal whitespace between the cells in the table.  Are you able to reduce/remove this please?

Screenshot 2023-07-28 at 08 21 20

IanMayo:
Thanks for the update @salony75way - but is it possible to remove the whitespace inside (and outside) the table as marked with the red lines?
image

salony75way:
> Thanks for the update @salony75way - but is it possible to remove the whitespace inside (and outside) the table as marked with the red lines? image

Hi @IanMayo, I have removed space from columns but its still not working with printed reports. We can only resolve this issue with changing page size from print preview or we can remove one column from report if possible.


I have changed page size from A4 to A3 and it looks like this:

Screenshot 2023-08-02 at 2 56 07 PM

IanMayo:
I _think_ it's caused by this `padding` CSS instruction:
image

salony75way:
> I _think_ it's caused by this `padding` CSS instruction: image
> 
> Note: the same padding is on the `th` header, too.

Ok let me check this

salony75way:
@IanMayo I have implemented this in code, can you please verify?
Screenshot 2023-08-02 at 5 11 44 PM

IanMayo:
This is what I see for the current deployment:
image

In the browser I cleared the CSS `padding` field:
image


and got this:
image

IanMayo:
That's the white-space fixed :-)

The title block has shrunk. Could you make the top two lines larger please?
image


Here is the form in `develop`:
image

They don't have to be as large as the above screenshot (from `develop`), but could they be a bit larger than they currently are?

IanMayo:
(still the title font-size issue, please @salony75way )

IanMayo (approved):

PR #584, opened 2023-07-31
audit log on destroy

### LINKED ISSUE #549: Need audit log entry for destruction complete
For dispatches, we create an event when the dispatch is sent out.

We should create an `AuditType` event for `DESTROY`, and create an audit event using it when the destruction is confirmed.

Note: when an item is added to a destruction, we create audit entries for the item and the destruction.  The item includes the `Name` of the item - so it can be linked to.  But the entry for the destruction omits this, so there isn't a clickable link:
image

Note 2: for the item entry, the details should include the reference of the destruction. For the destruction entry, it should include the reference of the item (like we do for dispatches)

salony75way:
fixes #549

### DIFF
--- a/src/components/ResourceHistory.tsx
+++ b/src/components/ResourceHistory.tsx
@@ -44,7 +44,12 @@ const ResourceHistoryModal = ({
     <Printable open={open} onClose={close}>
       <AuditList
         filter={{
-          activityType: AuditType.EDIT,
+          activityType: [
+            AuditType.EDIT,
+            AuditType.SENT,
+            AuditType.DESTROY,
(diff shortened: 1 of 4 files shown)


IanMayo:
@salony75way - I've just done this:
- create two destructions
- added 3 live items to the second destruction
- opened the second destruction
- pretended to print the certificate
- marked it as `destroyed`
- `Destroyed` audit events not shown for `item` or `destruction` (but I can see in the `Audit Log` that the correct events were generated.

So, I guess it must be some filter on the `History` page for `item` and `destruction`?

IanMayo (approved):

PR #586, opened 2023-07-31
added shortcut keys for form submission on item form

### LINKED ISSUE #557: Shortcuts for Save/Clone and Save/New
Item Create is by far the most frequent interaction with RCO.

We should create keyboard shortcuts for these two buttons.  This is so that a predominantly keyboard users can submit the form without having to tab to the bottom of the form first.

I suggest `<alt-c>` for `save/clone` and `<alt-n>` for `save/new`

I think the keyboard navigation should move to the cursor to the button, after which the user would press `enter` to activate it, rather then they keyboard short-cut fire that event.

The shortcuts should be added to the buttons, maybe in smaller text.

TahaKhanAbdalli:
fixes #557

### DIFF
--- a/src/resources/items/ItemForm/ItemFormToolbar.tsx
+++ b/src/resources/items/ItemForm/ItemFormToolbar.tsx
@@ -7,7 +7,7 @@ import { useContext, useEffect, useRef, useState } from 'react'
 import { SAVE_EVENT } from '../../../constants'
 import { transformProtectionValues } from '../../../utils/helper'
 import RemarksBox from '../../../components/RemarksBox'
-import { Button } from '@mui/material'
+import { Button, type ButtonBaseActions } from '@mui/material'
 import { DateTime } from 'luxon'
 import useVaultLocationAudit from '../../../hooks/useVaultLocationAudit'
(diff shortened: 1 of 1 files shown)


IanMayo (approved):

IanMayo (changes requested):
Thanks @TahaKhanAbdalli 

1. We need the short-cut to be on the button, please
2. <alt-c> and <alt-n> while creating a new item don't put the focus on the reelvant button for me

TahaKhanAbdalli:
> Ideally focus would shift to the button, like when we use tab-navigation (below). This would reassure the user which type of `create` is being performed.
> 
> https://i.gyazo.com/a36adb47dd24c3a6d8c7ec189e46c063.mp4

The focus is shifting to the button whose shortcut keys are pressed and then the user can hit enter to create that specific type of item.

IanMayo:
On ms-windows the shortcut keys work perfectly :-D  Maybe it's just macos where they don't.

We still need the keystroke labels on the buttons, please.

IanMayo (approved):

PR #598, opened 2023-08-01
auditlog summary in itemshow

### LINKED ISSUE #534: Pull summary of Audit Log onto ItemShow
Users are really interested in the history of an item, particularly the states it has been through.

So, they would like to promote the Audit Log for an `Item` to the `ItemShow` page.

Mockup below:
image

salony75way:
fixes #534

### DIFF
--- a/src/resources/items/ItemHistory.tsx
+++ b/src/resources/items/ItemHistory.tsx
@@ -1,0 +1,25 @@
+import { Datagrid, DateField, List, TextField } from 'react-admin'
+import { useParams } from 'react-router-dom'
+import SourceField from '../../components/SourceField'
+import * as constants from '../../constants'
+import FieldSet from '../../components/FieldSet'
+
+export default function ItemHistory(): React.ReactElement {
+  const { id } = useParams()
+  return (
+    <FieldSet title='History'>
+      <List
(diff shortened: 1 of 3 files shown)


IanMayo:
Thanks @salony75way .

Just one change please - can the audit items be in reverse chronological order please?

src/resources/items/ItemHistory.tsx
```diff
@@ -0,0 +1,25 @@
...
+import * as constants from '../../constants'
+import FieldSet from '../../components/FieldSet'
+
+export default function ItemHistory(): React.ReactElement {
+  const { id } = useParams()
+  return (
+    <FieldSet title='History'>
+      <List
+        resource={constants.R_AUDIT}
+        filter={{
```
IanMayo, line 13: @salony75way -  I think we also need to filter for entries with `resource: R_ITEMS`.  The id `1` could refer to audit entries for a range of resource types

IanMayo: Aah, one other filter, please. Could we exclude items where the activityType is `AuditType.EDIT`?

The full list of edits would be available under `History`, but not in this summary view.

IanMayo (approved):

PR #599, opened 2023-08-01
item list item columns bug fixed

### LINKED ISSUE #595: Issue showing `item` columns
Something weird is happening with the columns in ItemList that are further down the list.

I can't show `remarks`, `muster remarks` and `protection` at the same time.

_Maybe_ there is some confusion in the code regarding `remarks` and `muster remarks`?

TahaKhanAbdalli:
fixes #595

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -50,7 +50,7 @@ import { useLocation } from 'react-router-dom'
 
 const sort = (field = 'name'): SortPayload => ({ field, order: 'ASC' })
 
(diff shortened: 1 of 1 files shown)


IanMayo:
Could you give me an overview of what the issue was please @TahaKhanAbdalli ?

TahaKhanAbdalli:
> 

While conditionally rendering columns, the columns after the condition were causing problems. But when I shifted them to before the condition, they worked fine. So to make it more generic I declared them separately.

IanMayo (approved):

PR #600, opened 2023-08-01
always show action buttons

### LINKED ISSUE #548: Alternate strategy for `Action buttons` in ItemList
Users are unaware of what tasks they can do from the `ItemList` page.

This is an experiment.

Permanently show the grey actions toolbar, with all actions shown on it. Actions of relevance to the current selection will become enabled.  Note: since `Live Items` doesn't include items that are dispatched, it won't include `Return`.

Initially, this is just for the `Live Items` and `All Items` pages.

salony75way:
fixes #548

### DIFF
--- a/src/components/DatagridConfigurableWithShow.tsx
+++ b/src/components/DatagridConfigurableWithShow.tsx
@@ -53,8 +53,13 @@ export default function DatagridConfigurableWithShow(
       minHeight: '48px',
       height: '48px',
       position: 'relative',
+      transform: 'none',
+      justifyContent: 'flex-start',
       '& .RaBulkActionsToolbar-title': {
         display: selectedIds.length > 0 ? 'flex' : 'none'
+      },
+      '& .MuiBox-root': {
+        width: 'auto'
       }
     }
   }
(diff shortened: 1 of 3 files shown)


IanMayo:
Hello @salony75way  - I'm afraid I can't see what has changed.  Can you describe what I should see for this new feature please?

salony75way:
> Hello @salony75way - I'm afraid I can't see what has changed. Can you describe what I should see for this new feature please?

Hi @IanMayo For this, By default i am showing all action buttons to user but after selection on any items i am showing only enabled actions. Let me know if we need to show all action buttons here and need to disable if those are not matched to selection. 
Please check the attached video how its looking now with current implementation:

https://github.com/DeepBlueCLtd/RCO/assets/45573884/98376cff-b02d-4f05-aa40-6e7f2169dd5a

IanMayo:
Yes, that is the expected behaviour. When I first opened the `deployment`, I did see the full list. But, after a couple of selections I just see the previous behaviour (only show buttons relevant to current item, and none if nothing selected).

I can't get the `full list` to show by default at all now:
https://i.gyazo.com/0f62643501fc8cf1b4acacb66f9ff2cf.mp4

IanMayo:
Thanks @salony75way .   I think the list of actions should contain `loan return` and `return` too, please. So, that's 6 in total. Plus, when one or more items are selected, the invalid buttons are greyed out as disabled, while the others remain enabled.

IanMayo (approved):
Nice work. Let's give it a go.

PR #605, opened 2023-08-01
added remarks on change location on item list page

### LINKED ISSUE #604: Change Location popup should just contain active locations
From the `ItemList` it's possible to change the location of an item, using a popup dialog.

This dialog requires two changes:
- it should only list `active` vault locations
- it should have a text box for `remarks`.  The `remarks` text should be stored in the audit log for the change.

TahaKhanAbdalli:
fixes #604

### DIFF
--- a/src/components/FieldSet.tsx
+++ b/src/components/FieldSet.tsx
@@ -1,0 +1,30 @@
+import { Box, type BoxProps, Typography } from '@mui/material'
+
+interface Props extends BoxProps {
+  title: string
+}
+
+export default function FieldSet(props: Props): React.ReactElement {
+  const { title, children, ...rest } = props
+  return (
+    <Box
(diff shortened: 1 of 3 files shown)


IanMayo:
After I have changed the location for an item, the `Location` field for that row is empty.  

I've just checked - and this doesn't happen in `main`.

Hmm, it doesn't always happen.  I'll record screencast (later) if you can't reproduce it.

IanMayo:
> I couldn't reproduce. Can you please provide a screencast?

Not for 30 minutes or so.

IanMayo:
> I couldn't reproduce. Can you please provide a screencast?

I just experienced it again, but can't reproduce it for screencast.  Will ignore

IanMayo (approved):

PR #606, opened 2023-08-01
custom filter with nullable fields

### LINKED ISSUE #334: Custom filter doesn't work with nullable/empty fields
Custom filter doesn't work with nullable/empty fields

For example: If we have optional string field in database and we want to filter out items for empty/not-empty field with some toggle filter then custom filter (`_eq` and `_neq` operator) doesn't work in this case.

I _think_ this overlaps with #504

salony75way:
fixes #334

### DIFF
--- a/src/components/BooleanFilter.tsx
+++ b/src/components/BooleanFilter.tsx
@@ -17,20 +17,16 @@ export default function BooleanFilter<T extends RaRecord>(
 
   const { setFilters, displayedFilters, filterValues } = useListContext<T>()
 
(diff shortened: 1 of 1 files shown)


salony75way:
> Hello @salony75way - can you help me to test this please? What didn't work before, and works now?

Hi @IanMayo It is about `BooleanFilter`. It is working as usual but previously it was fetching all the items and then filter out based on the some conditions condition on FE like `{ dispatchJob !== undefined }` but now we are only fetching required data from BE with modifying and optimising filters like `{ dispatchJob_neq: undefined }`.

IanMayo (approved):

PR #607, opened 2023-08-02
increment consect on clone item

### LINKED ISSUE #542: Increment Consect counter on clone item
Consect/Pages will probably either contain a document reference (typically with a consecutive counter), a number of pages, or the acronym LEP.

When it's a document reference, frequently these will have a year number and a counter. These are samples of :
* 12/2023
* 12/23
* 040-23
* ARCH-32
* 1
* 01
* 12
* 035
* DDD 02
* 8888 <-  this is a system ID
* CON12/22
* EDR 12/22
* 1 of 8

I think we can match (and increment these patterns).  It's the **bold** item that would be incremented:
* **12**/2023
* **12**/23
* **040**-23
* **1** of 8
* CON**12**/22
* EDR **12**/22 <-- in the above cases, if there are two integers present, increment the first
* ARCH-**32**
* DDD **02**
* **1**
* **01**
* **12**
* **035** <- in the above cases, if one integer less than 1000 is present, increment it
* 8888 <-  this is a system ID, do not increment standalone numbers greater than 1000

We could do with a clever regexp to see if there is an integer in the string that is under 2000, and to increment it.

With this algorithm, we can modify `Save/Clone` to have a go at incrementing the Consect.  Users may end up writing over it, but it could that for 50 items, it _just-works_;

This snippet:
```
function incrementFormat(input) {
  return input.replace(/(\D*)\b(\d{1,3})\b/, (match, other, num) => {
    return other + (Number(num) + 1).toString().padStart(num.length, '0');
  });
}
```

(Test [here](https://jsfiddle.net/xh9aso5b/4/))

Generates these outcomes:
```
old: 12/2023
new: 13/2023
-----------
old: 12/23
new: 13/23
-----------
old: 040-23
new: 041-23
-----------
old: 1 of 8
new: 2 of 8
-----------
old: CON12/22
new: CON12/23
-----------
old: EDR 12/22
new: EDR 13/22
-----------
old: ARCH-32
new: ARCH-33
-----------
old: DDD 02
new: DDD 03
-----------
old: ANOTHER 07
new: ANOTHER 08
-----------
old: NEW-123
new: NEW-124
-----------
old: 1
new: 2
-----------
old: 01
new: 02
-----------
old: 12
new: 13
-----------
old: 035
new: 036
-----------
old: 0001
new: 0001
-----------
old: 0999
new: 0999
-----------
old: 1000
new: 1000
-----------
old: 8888
new: 8888
-----------
```

salony75way:
fixes #542

### DIFF
--- a/src/resources/items/ItemForm/ItemFormToolbar.tsx
+++ b/src/resources/items/ItemForm/ItemFormToolbar.tsx
@@ -183,7 +183,7 @@ const ItemFormToolbar = (props: Props): React.ReactElement => {
 
   function incrementFormat(input: string): string {
     return input.replace(
-      /(\D*)\b(\d{1,3})\b/,
+      /(.*?)(\b\d{1,3}\b)/,
       (_: string, other: string, num: string) => {
         return other + (Number(num) + 1).toString().padStart(num.length, '0')
       }
--- a/src/resources/items/ItemForm/ItemFormToolbar.tsx
+++ b/src/resources/items/ItemForm/ItemFormToolbar.tsx
@@ -75,7 +75,12 @@ interface Props {
 const ItemFormToolbar = (props: Props): React.ReactElement => {
   const { onSuccess, onSave, openRemarks, setOpenRemarks } = props
   const { notify } = useContext(NotificationContext)
-  const { reset, getValues, setValue } = useFormContext()
+  const {
+    reset,
+    getValues,
+    setValue,
+    formState: { errors }
+  } = useFormContext()
   const { id } = useParams()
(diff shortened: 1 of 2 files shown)


IanMayo:
I don't fully understand the logic in the code @salony75way .

I'd just like to double-check what is happening
* user provides data for new item
* user clicks on `Save/Clone` or `Save/New`
* RCO takes current value of `consec`, tries to increment it, and uses that new value for the _new_ item (not the one that was just saved).

IanMayo (approved):
Working perfectly!

PR #608, opened 2023-08-03
define bridging tables to RCOStore

### LINKED ISSUE #464: Define bridging tables
Our bridging tables are working, but we don't declare them in `types.d`.

Declare the bridging tables for batches and items with `cat-code`, `cat-handle`, `cat-cave`.

salony75way:
fixes #464

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -153,17 +153,17 @@ interface ItemHandling {
 interface BatchCode {
   id: number
   batch: number
-  catCode: number
+  catCode: CatCode['id']
 }
 interface BatchCave {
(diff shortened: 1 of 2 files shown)


IanMayo:
@salony75way I see the bridging tables have `number` as the type for `catCode`. These should actually be strings, because the `cat` tables have an integer id.

Better still, something like this:
```
interface ItemCode {
  id: number
  item: Item['id']
  catCode: StringReferenceItem['id']
}
```
Then it's quite clear what is happening.

IanMayo (approved):

PR #613, opened 2023-08-04
don't show dispatched items

### LINKED ISSUE #536: Live Items should not show dispatched/destroyed items
The `Live Items` page should not show dispatched/destroyed items.

It doesn't, when `Live Items` is clicked on.

But, if I click on `page refresh`, then dispatched/destroyed items are shown.

So, `Live Items` may display 75 items. But, if I click on `page refresh` then it can show 80 items.

Steps to reproduce:
- create destructions
- add some items
- complete destruction
- show `live items`
- search for an item reference that was destroyed.  It won't be shown
- press `Refresh`, the destroyed item is shown

salony75way:
fixes #536

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -537,6 +537,7 @@ export default function ItemList(
   ]
   return (
     <List
+      disableSyncWithLocation
       sx={sx}
       hasCreate={false}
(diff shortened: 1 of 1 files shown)


IanMayo:
fix works for me :-D

IanMayo (approved):

PR #617, opened 2023-08-07
stored the details of the users on login and logout

### LINKED ISSUE #615: Store user details for login/logout
For login/logout, we don't store the details of who has logged in/out.

So, store `userId` in `dataId` and `RESOURCE.R_USERS` in `resource`

TahaKhanAbdalli:
fixes #615

### DIFF
--- a/src/providers/authProvider/index.ts
+++ b/src/providers/authProvider/index.ts
@@ -58,8 +58,8 @@ const authProvider = (dataProvider: DataProvider): AuthProvider => {
           setToken(token, salt)
           await audit({
             activityType: AuditType.LOGIN,
-            resource: null,
-            dataId: null,
+            resource: constants.R_USERS,
+            dataId: user.id,
             subjectId: null,
             subjectResource: null
           })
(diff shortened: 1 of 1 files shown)


src/providers/authProvider/index.ts
```diff
@@ -72,10 +72,11 @@ const authProvider = (dataProvider: DataProvider): AuthProvider => {
       }
     },
     logout: async (): Promise<void> => {
+      const user = getUser()
       await audit({
         activityType: AuditType.LOGOUT,
-        resource: null,
-        dataId: null,
+        resource: constants.R_USERS,
+        dataId: user?.id ?? null,
```
IanMayo, line 79: @TahaKhanAbdalli - do you understand the occasions when `user` could be `null`?   I _presume_ we only offer `logout` for logged-in users.

TahaKhanAbdalli: Yes. If the user is not logged in then `dataId` would be `null` but if the user is `loggedIn` the user returned from `getUser` has a type of undefined or User, therefore, `user?.id`

IanMayo (approved):

PR #627, opened 2023-08-11
make start/end dates optional

salony75way:
fixes #626

PR #636, opened 2023-08-17
626 - start and end date on itemForm made optional

### LINKED ISSUE #626: Item Start/End date should be optional
Start and End date for items are optional in the legacy data.  But in the Item Form they are mandatory.

We should relax the front-end to make start/end dates optional.

In our `item` type they are described as optional (`string | null`).

I know this will require a database update - I'll do that in Taha's branch this afternoon.  **Done**.

TahaKhanAbdalli:
fixes #626


IanMayo:
We do have an existing PR for this #627 

Unsurprisingly, it's very similar to this. But, I suspect this PR will suffer from the same [issue](https://github.com/DeepBlueCLtd/RCO/pull/627#pullrequestreview-1580051879) experienced in #627 

Could you try to resolve that, in this branch please?

TahaKhanAbdalli:
It's working. If we don't provide the time, the form becomes invalid.

IanMayo:
While Date may be provided Time is optional, since it may not be present on the `item`.

I _guess_ that internally we just store it as 00:00 when time isn't provided.

But, the _issue_ in the branch is that on `Save/Clone` the `Finish-Date` of the current item is moved to the `Start-Date` of the cloned item. But, that doesn't work if only `date` is provided, but not `time`.

TahaKhanAbdalli:
If we put in `manual` values in date and not select from drop down (which selects time also), the start and end dates are empty in the form submission. 
image

IanMayo:
> If we put in `manual` values in date and not select from drop down (which selects time also), the start and end dates are empty in the form submission.

Wow!

IanMayo (approved):
Approved. Emergent bug relating to date/time not being tested captured in #626

PR #648, opened 2023-08-21
added env in github pages script

TahaKhanAbdalli:


TahaKhanAbdalli:
> Thanks Taha. There is no value in testing this by looking at the preview deployment in Heroku - since Heroku does set the correct env var. So, I'll just merge it, and see how GitHub pages works with it. Ok?

Yes, please

IanMayo (approved):

PR #649, opened 2023-08-21
fixed fresh item containing lots of codes, caves 

### LINKED ISSUE #644: Fresh `item` contains lots of codes, caves, handles.
When I add an item to a batch, the new item has lots of codes, caves, handles pre-assigned.  They should be empty.

screencast

TahaKhanAbdalli:
fixes #644

### DIFF
--- a/src/components/ProtectionBlockInputs.tsx
+++ b/src/components/ProtectionBlockInputs.tsx
@@ -4,7 +4,7 @@ import { Typography } from '@mui/material'
 import { Box } from '@mui/system'
 import SourceInput from './SourceInput'
 import ProtectionRefInput from './ProtectionRefInput'
-import { useDataProvider, type RaRecord } from 'react-admin'
+import { type RaRecord } from 'react-admin'
 import { useFormContext } from 'react-hook-form'
 import { useConfigData } from '../utils/useConfigData'
 import { useEffect, useState } from 'react'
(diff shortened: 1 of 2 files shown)


src/components/ProtectionBlockInputs.tsx
```diff
@@ -46,27 +45,6 @@ export default function ProtectionBlockInputs<
     })
   }
```
IanMayo, line 47: @TahaKhanAbdalli - can you explain why we've deleted the following block of code please?

IanMayo (approved):
Tested and working

IanMayo (changes requested):
Aaah, there is still a pending debug line in the second file.

Could you also add a comment explaining why we're checking for mock env variable in the second file please @TahaKhanAbdalli ?

TahaKhanAbdalli:
> Aaah, there is still a pending debug line in the second file.
> 
> Could you also add a comment explaining why we're checking for mock env variable in the second file please @TahaKhanAbdalli ?

We are checking for mock data because `mock` accepts `undefined` and `SQLite` accepts `null`

src/components/ProtectionBlockInputs.tsx
TahaKhanAbdalli: This is `pMarking`, that we discussed with salony and it needed to be removed becuase our db has protectiveMarking not pMarkin.

IanMayo:
> We are checking for mock data because `mock` accepts `undefined` and `SQLite` accepts `null`

Aah, ok. It will be good when [this](issue us resolved), and we just have `null`.

IanMayo (approved):

PR #652, opened 2023-08-21
destroy item bug fixed

### LINKED ISSUE #650: Bug - for destroy items, only show incomplete destructions
Here is the process to destroy items:
- create `destructions`
- add `items` to `destruction`
- print off `destruction report`
- click on `finalise destruction`

After the last item, it should not be possible to add more items.  So, in the popup box for `Add item to destruction`, it should only offer destructions that do not have a `finalisedAt` date.

I _think_ this may be specific to `SQLite`, I _think_ it's working for mock.

TahaKhanAbdalli:
fixes #650 

Will only work for `SQLite` as `SQLite` uses `null` whereas `mock` server uses `undefined`


IanMayo (approved):
fix looks valid

PR #653, opened 2023-08-21
destruction page crashing fixed after printing

### LINKED ISSUE #643: Create `Destruction` fails for SQLite
I've created a new destruction.  SQLite has given it a a row id of `1`:
image

But, the `Show` page is trying to retrieve destruction with row id of `0`, and it is failing.
image

When I view the `Destructions` table, I can view individual destruction jobs.

Ok, there are some other dodgy bits around destructions.  I can go to `print` the form, then cancel, then the form falls over.

TahaKhanAbdalli:
fixes #643 

Reference: [better-sqlite3 docs](https://github.com/WiseLibs/better-sqlite3/blob/HEAD/docs/api.md#runbindparameters---object)

image

### DIFF
--- a/src/providers/dataProvider/index.ts
+++ b/src/providers/dataProvider/index.ts
@@ -264,7 +264,7 @@ export const dataProvider = (apiUrl: string): DataProvider => ({
 
     return await axios.put(url, { fields: editData }).then((response) => {
       return {
-        data: { id: response.data.data.lastInsertRowid, ...params.data }
+        data: { id: params.id, ...params.data }
       }
     })
   },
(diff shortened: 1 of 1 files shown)


src/providers/dataProvider/index.ts
```diff
@@ -262,9 +262,9 @@ export const dataProvider = (apiUrl: string): DataProvider => ({
     // remove the id property
     const { id, ...editData } = params.data
 
-    return await axios.put(url, { fields: editData }).then((response) => {
+    return await axios.put(url, { fields: editData }).then(() => {
       return {
-        data: { id: response.data.data.lastInsertRowid, ...params.data }
+        data: { id: params.id, ...params.data }
```
IanMayo, line 267: haven't we extracted `id` as a `const` in line 263 @TahaKhanAbdalli ?

TahaKhanAbdalli: No, the `id` is `undefined` as we are only sending `reportPrintedAt` prop to update function.

IanMayo (approved):

PR #654, opened 2023-08-21
start and end dates made required on item form

### LINKED ISSUE #640: Improve validation of start/end date
@TahaKhanAbdalli  said:

> The dateTimeInput component doesn't trigger onChange until filled completely which means validation won't run if it doesn't detect anything

That means it is possible for a user to add an item with a partial date, but the form validator won't check it, and the item will be saved without a date value.

### LINKED ISSUE #647: Bug - start/end dates only verified if time present
For the time control we're using, the form only verifies the data if a full date/time is entered.

If date but not time is entered, then no verification is performed, a not date-time is even stored for that field. It's empty.

Related to #640

TahaKhanAbdalli:
fixes #647 
fixes #640


IanMayo:
Sorry - the language of dates and times is confusing.

We need the `startDate` and `endDate` to be optional, since some items are not be related to a time period.  Our issue is that the `time` element of the `date-time` was optional, and the field got ignored if a time wasn't present.

So, while the whole of `startDate` can be optional, if a value is provided then we need it to contain a value for `time` (hh:mm)

TahaKhanAbdalli:
> Sorry - the language of dates and times is confusing.
> 
> We need the `startDate` and `endDate` to be optional, since some items are not be related to a time period. Our issue is that the `time` element of the `date-time` was optional, and the field got ignored if a time wasn't present.
> 
> So, while the whole of `startDate` can be optional, if a value is provided then we need it to contain a value for `time` (hh:mm)

Yes. But we have no way of knowing if the user has entered the date or user hasn't entered anything because the input element of html of type `datetime-local` triggers `onChange` only when the whol value is filled. We can't force the user to also enter `time` if date is entered.

IanMayo:
Ok.  So, _maybe_ the fix to this issue is to find a react date-time component that behaves in the way we require.  Could that work?

TahaKhanAbdalli:
> Ok. So, _maybe_ the fix to this issue is to find a react date-time component that behaves in the way we require. Could that work?

They all, under the hood, use input element of html.

IanMayo:
> They all, under the hood, use input element of html.

Sure, but the events they trigger, and the constraints on the text can be different.  I don't think our expectations are _that_ unreasonable.

An alternate issue would be for the user to toggle a `not date related` button, which disables the two date controls, otherwise they are compulsory.

Hmm, but this issue `isn't the biggest alligator at the canoe`.  I suggest we close this PR, and return to the issue another time.

TahaKhanAbdalli:
Sure! Let's come back to it later.

IanMayo:
Closing this. We aren't able to make start/end date fields optional.

IanMayo (approved):

PR #656, opened 2023-08-21
optional types converted to null

### LINKED ISSUE #504: Operators required for REST backend
We need to handle filtering operators in `soul` as we are using custom filtering operators in our `react-admin` application. Following is the list of the spotted filter operators:

- [x] _eq
- [x] _neq
- [x] ~~q (This versatile search filter allows matching any record within the SQLite table, regardless of the field, providing comprehensive search capabilities for your data.)~~. @TahaKhanAbdalli reported that `_search` fill requirements of `q`.
- [x] lte
- [x] gte
- [ ] ~~comparison with undefined value e.g. `filter={{ receiptReceived: undefined }}`~~  **Note: hopefully we can remove instances of `undefined`**. See [below](https://github.com/DeepBlueCLtd/RCO/issues/504#issuecomment-1649396459)
- [x] comparison with null value e.g. `filter={{ receiptReceived: null }}` [issue created](https://github.com/thevahidal/soul/issues/105)

TahaKhanAbdalli:
Types are converted to `null` but `mock` localForage still needs `undefined`.

Fixes #504


IanMayo:
Ok, @TahaKhanAbdalli - could you have a look at my changes please?  

I've tried to ensure full JS object gets created, with relevant `null` fields. I'm also trying to handle providing the correct filter operator.

TahaKhanAbdalli:
> Ok, @TahaKhanAbdalli - could you have a look at my changes please?
> 
> I've tried to ensure full JS object gets created, with relevant `null` fields. I'm also trying to handle providing the correct filter operator.

Yes. It seems ok and it's working as expected.

PR #659, opened 2023-08-22
location field bug fixed on item list page

### LINKED ISSUE #657: `Location` column in `ItemList` not filled for SQLite
For the `SQLite` backend the `LocationField` component does not provide text to render.

image

On first inspection, it _looks-like_ `LocationField` receives an empty props once opened.

We should _probably_ only update the `setUsers` and `setVaultLocations` if the props contains `non-empty` datasets.

Note: the data doesn't come in as arrays, but as indexed records. This is because the id values may not be continuous, so data can't be accessed by array index.

TahaKhanAbdalli:
fixes #657


IanMayo (approved):
Tested and working

PR #660, opened 2023-08-22
recent loan item fixed on welcome page 

### LINKED ISSUE #658: Welcome Page - Items on Loan
The welcome page includes a listing of items on loan.  Note: I've seen this failing in SQLite, it probably works in `mock` due to the smaller volume of data.

It is empty in SQLite, despite there being items on loan:

image

image

I think the cause is that it is looking for loaned items in the first page of returned results, then collating a list of unique users with items on loan.  But, there are not any in the first page.
Instead we need to request a list of items on loan, then generate the unique list of user names from that data.

TahaKhanAbdalli:
and support for on loan filter added for `mock` data
fixes #658


IanMayo (changes requested):
Thanks @TahaKhanAbdalli   The correct items are shown on the welcome page, and when I click on a user name, the ItemList page opens.  It seems to have a valid filter in the URL, but the filter isn't applied: all items are visible.

IanMayo (approved):
Working fine.

PR #662, opened 2023-08-22
introduced enduring projects

### LINKED ISSUE #637: Introduce "Enduring" Projects
Sometimes RCO users use the Project Name as a way of recording the batch as supporting some kind of business process (such as "RECYCLING").

This results in "RECYCLING" getting added as a project name lots and lots of times.

Instead, we will create some "ENDURING" projects.  These have very large start/end time range (1980-2060). They have a title like `RECYCLING (ENDURING)`.  The projects table will gain an `enduring` boolean` attribute.

For generated data, before generating the sequential project entries insert 3 enduring projects:
- `RECYCLING (ENDURING)`
- `DESTRUCTION (ENDURING)`
- `TEST MEDIA (ENDURING)`

This will require:
* [x] modifications to `types.d`
* [x] modifications to SQLite
* [x] modifications to `generate-data` to prepend these enduring tasks
* [x] modifications to `Create Batch` to prepend the `enduring` items before the most recent 10 `non-enduring` projects (see below)
* [x] modifications to `Edit Project` to allow the `enduring` boolean flag to be edited

image

TahaKhanAbdalli:
fixes #637


IanMayo (approved):
Tested and working

PR #664, opened 2023-08-23
runtime console errors fixed on destruction list and show pages

### LINKED ISSUE #663: Runtime console errors
There are a few runtime console errors.

One is about `hasCreate` is not a react attribute, and another is that we can't show destruction form while something else is updating.

We should identify, and fix, these.

TahaKhanAbdalli:
fixes #663


IanMayo:
Thanks @TahaKhanAbdalli 

Note: there is still a console error when I open `Vault Location List`.
image

Also one in `User List`:
image

One on `Create Destruction`:
image

IanMayo (approved):

PR #665, opened 2023-08-23
switching to support for types in react-admin fields

### LINKED ISSUE #395: Switch to types
React-Admin components can now take a type parameter, to ensure the correct field(s) are selected:
https://marmelab.com/blog/2023/06/13/react-admin-june-2023-update.html#generic-field-components

But, this doesn't seem to work for `interfaces`, not for `types`.

@salony75way to confirm if this would be a backward step or not.

So amend `types.d.ts` so we use types and not interfaces.  For inheritance, we need to use this trick:https://stackoverflow.com/a/53636674/92441

Then update react-admin to newest stable version.

TahaKhanAbdalli:
fixes #395


src/resources/audit/AuditList.tsx
```diff
@@ -191,7 +195,7 @@ export default function AuditList({
           label='Subject'
           render={(record) => {
             return (
-              <SourceField
```
IanMayo, line 194: @TahaKhanAbdalli - while you are working in this file, can I suggest you use the resourcesKeyRef index in the next SourceField like you do in the previous SourceField?  
That's more versatile than the hard-coded R_ITEMS check.

IanMayo:
Hello @TahaKhanAbdalli  - I wouldn't expect this issue to be related to the changes in this PR, but I'm unable to create a new `Destruction`.  I'm in `dev:mock`:

image

IanMayo (approved):

PR #668, opened 2023-09-05
password can only be set but not viewed

### LINKED ISSUE #630: Cannot view password for other users. Can only set new password (logged)
We can't let users see the password for other users (encrypted or not).

But, we do need to allow passwords to be updated, in case of lost password.

Document this in authentication diagram, implement in UI.

TahaKhanAbdalli:
fixes #630


IanMayo (approved):
Tested and working

PR #676, opened 2023-09-07
95 batch year to number

IanMayo:
Supports #95


IanMayo:
Drop this, it's not working.  Let Taha have a go :-D

PR #680, opened 2023-09-07
save button enabled by default

### LINKED ISSUE #674: Bug - forced to add remarks for Destruction
As a user I can only create a destruction once Remarks have been provided.

They should be optional.  I've checked the `yup` schema, and it looks ok.
image

TahaKhanAbdalli:
fixes #674


IanMayo (approved):
tested and working

PR #682, opened 2023-09-08
yearOfReceipt in batch converted to number

### LINKED ISSUE #677: yearOfReceipt should be number
In Batch we currently store `yearOfReceipt` as a string.

It will always be a number, so we should store it as a number.

* [x] update `types.d.ts`
* [x] update front-end code as necessary
* [x] test `create new batch` in  `dev:mock`
* [x] verify front-end unit tests run
* [x] modify backend (`RCO2.sqlite` and `full.sql`)
* [x] test `create new batch` in `dev`

Note: this is part of #95

TahaKhanAbdalli:
fixes #677


IanMayo (approved):
tested and working

PR #684, opened 2023-09-08
338 filters added on dispatch and destruction list

### LINKED ISSUE #338: Dispatch list filters
Toggle filter doesn't work with nullable fields.

- Not dispatched (dispatchedAt is undefined/null),
- Receipt not received (receiptReceived  is undefined/null)

**Updated objectve**:
* [ ] introduce "Filters" for dispatches and destructions
* [ ] Introduce "Dispatched", "Not dispatched", "Receipt note pending" toggle filters for dispatches

Related issue: #334

TahaKhanAbdalli:
fixes #338


IanMayo:
One change: the filter for `pending receipt note` is that `dispatchedAt` is non-null, and `receiptReceived` is null.

So, we only show items that have been sent, but for which receipt note is still awaited.

src/components/NullUndefinedFilter.tsx
```diff
@@ -0,0 +1,27 @@
...
+  source: string
+}
+
+const NullUndefinedFilter = (props: Props): React.ReactElement => {
+  const { label, source } = props
+  const { setFilters, displayedFilters, filterValues } = useListContext()
+
+  useEffect(() => {
+    setFilters(
+      {
```
IanMayo, line 16: we seem to insert `filterValues` twice, below.

PR #685, opened 2023-09-11
items can be removed from destruction

### LINKED ISSUE #681: Need to be able to remove items from a destruction job
We can add items to a destruction job.

If the destruction job hasn't been completed, then we also need to be able to remove them from the destruction job.

See how we do this from `dispatchShow` for dispatches.

TahaKhanAbdalli:
fixes #681


IanMayo (approved):
tested and working

PR #689, opened 2023-09-11
login using staffNumber instead of username

### LINKED ISSUE #632: Login using staff id, not user name
~~The legacy data has first-name and last-name fields, not username.~~

No - since the two fields are never used separately, combine them into one.

Login is also via staff id, not user name.

* [ ] ~~change `types.d`. to match above~~
* [ ] ~~change SQLite backend to match above~~
* [ ] change login screen/process to staff-id, not username.
* [ ] change `User` form to refer to `Staff ID`, not `Staff Number`

TahaKhanAbdalli:
fixes #632


IanMayo:
Thanks Taha - this won't be needed for long, but could you put the staff ids on the welcome page?
It's currently showing name (`ian` and `jason`).
Thx.

IanMayo:
Oh, and can you describe what has changed in `RCO2.sqlite` please?
(the PR doesn't show me, but I wasn't expecting any changes)

TahaKhanAbdalli:
> Oh, and can you describe what has changed in `RCO2.sqlite` please? (the PR doesn't show me, but I wasn't expecting any changes)

This change should have occured in the previous PR where we removed the salt from the user table, but I forgot there. So, I removed the `salt` from the `user` table in this `RCO2.sqlite`

IanMayo (approved):
working fine

PR #694, opened 2023-09-12
project list fetching filter changed on batch form

### LINKED ISSUE #691: Strategy for high numbers of projects
There are 3500 projects in the legacy data.

The ProjectList view is paged, so react-admin will only ever gather 100 records at a time.

For most data, the `project` is just a reference field - so `react-admin` pulls in projects for relevant ids as required.

But, for `Add Batch` - the drop-down needs to offer all projects.  Well, it doesn't actually need all of them.  It needs all enduring projects, plus any with a finish date in the last two years that are marked as `active` (new flag introduced in #692)

If possible, I'd like us to amend the `BatchForm`  so that the `project` control just pulls from this shorter set of projects.

If we can't do that, I think that _unfortunately_ we'll need to create `All Projects` and `Live Projects` resources.

Depends on #692

TahaKhanAbdalli:
fixes #691


IanMayo (approved):

IanMayo:
tested and working

PR #699, opened 2023-09-13
rich item view implemented

### LINKED ISSUE #389: Allow `item` search by `platform` and `project`

TahaKhanAbdalli:
fixes #389


TahaKhanAbdalli:
@IanMayo still need to fix synchronization of `item` and `richItem` on `MOCK` server

IanMayo:
Thanks Taha, but I _think_ that RICH_ITEMS should only be referred to when declaring the `Live Items` and `All items` listings.  I'm pretty sure the others should just refer to `ITEMS`.

TahaKhanAbdalli:
> Thanks Taha, but I _think_ that RICH_ITEMS should only be referred to when declaring the `Live Items` and `All items` listings. I'm pretty sure the others should just refer to `ITEMS`.

On some instances, we need to navigate to `richItem` and for `Edit` and `Create` we need to explicitly tell that we are referrring to `Item` resource because by default react-admin refers to the resource using `url`

IanMayo (approved):
tested and working

PR #702, opened 2023-09-14
active dropped for users

### LINKED ISSUE #698: Drop `active` for users
We should not use `active` flag for users.

We should check if the `endDate` is:
- non-null
- in the past

We'll have to update the database for this, too

We use this status to control which people can have items `loaned` to them.

TahaKhanAbdalli:
fixes #698


IanMayo:
Oops, I forgot. We'll need to be able to edit the Departure Date in `UserEdit`, please.

TahaKhanAbdalli:
> Oh and a minor (unrelated) issue.
> 
> In the drop-down list of users for `loan to`, could the users be in ascending order of staff number, with user-name in brackets after it?. We also need to only offer users where the departure date is either null, or a date in the future.

How can `departedDate` be in the future?

IanMayo:
> How can `departedDate` be in the future?

🤣 🤣 🤣 

I know, I know, that's just the state of the legacy data.  It is, what it is.  For people that are still here, they mostly have a `departed date` in 2028.  I can only guess it's so that unless the date gets updated, they will automatically lose access in 5 years.

IanMayo (approved):
tested and working

PR #703, opened 2023-09-15
only logged in users can view data

### LINKED ISSUE #697: Must be logged in to view RCO
If not logged in, users only have a read-only view of data.

Clients have asked that we switch strategy - so that non-logged in users can only see the login page.  They must log in to view RCO

TahaKhanAbdalli:
fixes #697


src/hooks/tests/protected-route.test.tsx
```diff
@@ -16,18 +16,20 @@ const resourceRoutes: ResourceRoutes = {
...
-      constants.R_ITEMS,
-      resourceRoutes
-    )
-    expect(list).toBeDefined()
-    expect(show).toBeDefined()
-    expect(create).toBeUndefined()
-    expect(edit).toBeUndefined()
-  })
+  // no longer required as unauthorized users can't see any data
+
```
IanMayo, line 20: Hello @TahaKhanAbdalli - shouldn't we check these objects are _not_ defined?

TahaKhanAbdalli: Since now there's no `user` role, therefore, we can't check these objects for any undefined role.

IanMayo: Aah, that makes sense :-)

IanMayo (approved):
Tested and working

PR #707, opened 2023-09-15
minor bug fix and welcome page receipt notes filter fixed

### LINKED ISSUE #705: On Welcome Page "Receipt Notes" should just be listed for dispatched items
We have `Pending Receipt Notes` on the welcome page.

This currently lists dispatches that have not had a receipt note.  But, it should list dispatches that have been sent (`dispatchedAt is non-null`, but without a receipt note (`receiptReceived is null`).

Further, dispatches should be shown in `Red` if `receiptReceived is null` and either:
- a hastener has been sent, but it was more than 4 weeks ago
- a hastener has not been sent, and the `dispatchedAt` is more than 4 weeks ago

TahaKhanAbdalli:
fixes #705

### DIFF
--- a/src/components/PendingReceiptNotes.tsx
+++ b/src/components/PendingReceiptNotes.tsx
@@ -1,34 +1,38 @@
 import { DateTime } from 'luxon'
 import Recent from './Recent'
 import * as constants from '../constants'
-import { DateField, type Identifier, useGetList, useTheme } from 'react-admin'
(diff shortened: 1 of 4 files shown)


IanMayo:
One linter fix, please @TahaKhanAbdalli 
image

IanMayo:
Thanks Taha, will do some testing when at home - so I can play with the SQLite data.  Then will accept it :-)

IanMayo (approved):
tested and working

PR #711, opened 2023-09-18
vault not showing on dispatch fixed

### LINKED ISSUE #710: Not showing VAULT for dispatch using old (legacy) vault
I'm usingt the SQLite backend.

We have 2 dispatches in the test data.  One has `vault` of `VAULT` the other has `vault` of `LEGACY`.

When I view the dispatch for `LEGACY`, the value for `vault` isn't shown, and a console error is shown.
image

image

TahaKhanAbdalli:
fixes #710

### DIFF
--- a/src/resources/batches/BatchForm.tsx
+++ b/src/resources/batches/BatchForm.tsx
@@ -39,18 +39,18 @@ interface Props {
   source: string
   inputProps?: SelectInputProps | TextInputProps
   active?: boolean
+  show?: boolean
 }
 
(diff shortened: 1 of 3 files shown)


IanMayo:
Hello @TahaKhanAbdalli - I've just realised we don't show `Vault` for destructions, but we should.  The field is present in the data, and users may need to check it for old (legacy) data.

Could you please add `vault` to the the form(s) for destructions, just like you have for dispatches please?

PR #712, opened 2023-09-18
709 update auth

### LINKED ISSUE #709: Updated authentication
We have an updated strategy for authentication:
https://docs.google.com/drawings/d/1y9mCS0RH38RIAJfPraqeqblCDTNOZwC2BtLH53mpzVI/edit

We should implement it.

TahaKhanAbdalli:
fixes #709


IanMayo:
Working well so far, @TahaKhanAbdalli . One issue - after the new user has changed their password, we should create an audit log for the event.

`activityType` would be `Edit`.  `activityDetails` would be `Password assigned`. `resource` would be `user`.

IanMayo:
In Slack we've agreed that we do not need to SALT the user object before it get encrypted in the browser.

IanMayo:
It's only a minor issue, but would you mind switching this loop to `for ... of` please?
image

IanMayo:
> It's only a minor issue, but would you mind switching this loop to `for ... of` please? 
Done.

IanMayo:
Have tested that user without password gets invited to enter it on login.
Have verified that changing password is audited.

IanMayo (approved):

PR #715, opened 2023-09-20
ip address added in audit entries

### LINKED ISSUE #714: Include IP address in audit log?
Is it possible to include the client IP address in the audit log?

This would enhance security, if we were able to do it.

Will require change(s) to audit log in backend.  But, field must be optional, since I can't provide it for legacy audit log entries.

TahaKhanAbdalli:
fixes #714


src/utils/helper.ts
```diff
@@ -17,3 +18,24 @@ export const checkIfUserIsActive = (user: User): boolean => {
   }
   return true
 }
+
+const getIp = async (): Promise<string | undefined> => {
+  try {
+    const data = await axios.get('https://api.ipify.org?format=json')
```
IanMayo, line 24: Sorry @TahaKhanAbdalli - the clients will be running this on a network without an Internet connection, so we can't make the above call.

TahaKhanAbdalli: Then, there's no easy way to fetch the ip with vanilla js

IanMayo: Ok. Could you investigate how to do it with node please?  `Soul` supports _extensions_, which are blocks of server-side NodeJS code.  Maybe we can develop a Soul extension to give us the caller's IP address.

src/utils/helper.ts
```diff
@@ -17,3 +18,24 @@ export const checkIfUserIsActive = (user: User): boolean => {
   }
   return true
 }
+
+const getIp = async (): Promise<string | undefined> => {
+  try {
+    const data = await axios.get('http://localhost:8000/api/ip')
```
IanMayo, line 24: I'm pretty sure we need to take the port from the env variable

IanMayo:
Thanks @TahaKhanAbdalli - we also need to handle the populating the $EXTENSION variable.  _Maybe_ in the `.env`?

src/utils/helper.ts
```diff
@@ -17,3 +18,26 @@ export const checkIfUserIsActive = (user: User): boolean => {
...
 }
+
+const getIp = async (): Promise<string | undefined> => {
+  try {
+    const data = await axios.get(
+      `${process.env.VITE_API_URL ?? 'http://localhost:8000/api'}/ip`
+    )
+    return data.data?.ip
+  } catch (error) {
+    return undefined
```
IanMayo, line 29: should we log any errors to the console @TahaKhanAbdalli ?

IanMayo (approved):

PR #722, opened 2023-09-21
vault location made optional on item

### LINKED ISSUE #718: vaultLocation in item needs to be optional
We have `items` for which the `vaultLocation` is null.

Don't know why, but they're there.  We need to make the change in types.d and SQLite. We should also modify the mock data generation so that we have some items with a null for `vaultLocation`, and check that the UI works ok.

TahaKhanAbdalli:
fixes #718


IanMayo (approved):
make item.vaultLocation optional

PR #726, opened 2023-09-28
rco served from soul extension

### LINKED ISSUE #725: Tunneling approach
We have issue on client network.  While the browser can access port 8080 on the VM (actually mapped to port 80), the network can't be configured to allow access port to 8000.

The above was learned at the same time as learning that we can't run on `localhost` URL, since it's the device the user is running from which is what is used for the URL, so RCO is looking for Soul on the user device, not the server.  

We were able to correct the `.env` URL to point to `VAL:8000`, and it worked for the sys admin guy (for whom all ports are open), but not for a plain user, for whom only 80 is open.

So, we RCO to use just one port for both RA and Soul.  I see two options:
1. We modify our _thin_ node hosting process to support a URL which acts as a proxy to getting the data from localhost:8000.  So, `http://VAL/Soul/api/tables/user` would get mapped to `http://localhost:8000/api/tables/user`.
2. We modify Soul installation to host RCO as an extension, with RCO being URL `http://VAL/api/RCO` .  Then when RCO needs to make calls, they will be to `http://VAL/api/tables/user`, etc.  We also drop the local _thin_ hosting process, and sysadmin maps 80 port to 8000.

TahaKhanAbdalli:
fixes #725

### DIFF
--- a/_extensions/api.js
+++ b/_extensions/api.js
@@ -1,3 +1,6 @@
+const path = require('path')
+const express = require('express')
+
 const getIp = {
   method: 'GET',
   path: '/api/ip',
@@ -6,6 +9,17 @@ const getIp = {
   }
 }
 
+const rcoClient = {
+  method: 'GET',
+  path: '/',
+  handler: (req, res) => {
+    const clientPath = path.join(__dirname, '../dist', 'index.html')
+    res.app.use(express.static(path.join(__dirname, '../dist')))
+    res.sendFile(path.join(clientPath))
+  }
+}
+
 module.exports = {
-  getIp
+  getIp,
+  rcoClient
 }


TahaKhanAbdalli:
> Thanks Taha - don't we need an updated command in `package.json` to run RCO from the new extension?
> 
> Hmm, is it possible to check for the existence of `\dist` folder - and throw a `Please run "yarn build" warning if it isn't found?

Should we not run `yarn build` before serving rather than throwing error and asking user to run the command?

IanMayo:
> > Thanks Taha - don't we need an updated command in `package.json` to run RCO from the new extension?
> > Hmm, is it possible to check for the existence of `\dist` folder - and throw a `Please run "yarn build" warning if it isn't found?
> 
> Should we not run `yarn build` before serving rather than throwing error and asking user to run the command? @IanMayo

Yes - `yarn build` is in the installation script.  It's just that we could robustly handle this quite understandable installation error.

TahaKhanAbdalli:
And for a new script, we don't need that because when we run `yarn serve:soul`, it's serving `RCO` on port `8000`

IanMayo:
> And for a new script, we don't need that because when we run `yarn serve:soul`, it's serving `RCO` on port `8000`

Sure, but doesn't `rcoClient` need to be included in the URL?

TahaKhanAbdalli:
> > And for a new script, we don't need that because when we run `yarn serve:soul`, it's serving `RCO` on port `8000`
> 
> Sure, but doesn't `rcoClient` need to be included in the URL?

That will not work because we are using middleware to serve our static content. And the middleware doesn't run until we hit the path specified in the extension. So, even if we add `/rcoClient` it will not serve other static content `.js` files etc.

TahaKhanAbdalli:
> > And for a new script, we don't need that because when we run `yarn serve:soul`, it's serving `RCO` on port `8000`
> 
> Sure, but doesn't `rcoClient` need to be included in the URL?

That is not working because we are bounding a middleware if a request to specific path is made and the rest of the static content is served from `/` path. Event if I try to bound the middleware on the same path we are serving our content from, it doesn't load files other than `index.html`.

TahaKhanAbdalli:
The latest approach of checking dist folder first will only work in bash environment.

IanMayo:
Is it an option to check it in nodeJS code, at the same time as declaring the `rcoClient` extension?

TahaKhanAbdalli:
> Is it an option to check it in nodeJS code, at the same time as declaring the `rcoClient` extension?

No we can't do that because the `error` will not be thrown unless the `rcoClient` path is not hit.

IanMayo:
I don't know why this is specific to this branch, but it doesn't occur on `main`.

When I view `live items`, the `Location` column is filled.  When I view an item, then go back, the `Location` column is empty. 

Sometimes the `Location` is empty after switching between tabs.

screencast

IanMayo:
Could this PR also please include an update to the root-level README please?

For `dev`, it's probably still worth running two separate processes.  For `production` we should use this single-command strategy.

IanMayo (approved):

PR #737, opened 2023-10-04
active filter added on reference data list

### LINKED ISSUE #733: Reference tables - allow filter on Active
On the `Reference Data` page we have `Media Type` (and other) views.

This view should allow users to filter the data to only show `Active` entries.

This is valid for most lookup lists, though not all.

TahaKhanAbdalli:
fixes #733

### DIFF
--- a/src/components/ReferenceDataList.tsx
+++ b/src/components/ReferenceDataList.tsx
@@ -7,16 +7,20 @@ import {
   List,
   TopToolbar,
   BooleanField,
-  TextField
+  TextField,
+  FilterButton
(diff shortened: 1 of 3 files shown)


IanMayo:
Oh, and can we have an `Active` filter on `Vault Locations` too, please?  Actually, it should be `on` by default.

TahaKhanAbdalli:
> Oh, and can we have an `Active` filter on `Vault Locations` too, please? Actually, it should be `on` by default.

Done

IanMayo (approved):

Issue #732, opened 2023-10-03
Can't edit existing vault location


IanMayo:
I don't have access to the bulk dataset from here, so I can't run it against bulk - the issue doesn't happen in the normal dataset.

But, here is the issue:
- I view an item
- I see the vaultLocation (P45)
- I click on `Edit`
- The `vaultLocation isn't set to P45 - possibly because it has only retrieved first 25 rows of vaultLocation

If I type `P45` in the drop-down doesn't do a round trip to see if there is a location called `P45`.  I _Thought_ react-admin did this in other forms.

TahaKhanAbdalli:
This is a `system-wide` issue, we have done it in many palces either by `hard-coding` number of rows or by using default 25 rows.

IanMayo:
I _Think_ I have seen in other places, that when I (as a user) starts to type, RA sends a wildcard search to the backend to retrieve matching items.

IanMayo:
This is done.

PR #739, opened 2023-10-05
showing only active on add and all for edit

### LINKED ISSUE #738: Only offer `active` on `Add`, but all for `Edit`
Legacy data will contain values that are no longer `active`.

So, when the record is in `Show` - RA pulls out the FKs and displays the label/name for the item.

When we `Add` an item, we should only allow `Active` items.

But, when we `Edit` an item, we should allow the full list of values to be displayed - so that the current value is shown.  Some lists are really long when we include all items (esp media-types, platforms, projects, users).  In these cases, I _think_ we need to use an autocomplete box, so RA isn't trying to show a list of 1000 items.

These are the controls with long lists:
**Item**: media-type, vault-location
**Batch**: platform, project

Aah, when we show the full list (active and non-active values), we should render ` (Legacy)` after items that are no longer active.

TahaKhanAbdalli:
fixes #738 #732


IanMayo:
One issue. When I add a new batch, I now have to assign organisation and department. Our logic has been that if there is just one value available, then it gets assigned by default (and the control is effectively read-only/disabled)

IanMayo:
* We aren't showing `(Legacy)` for inactive platforms on `Edit Batch`
* I can't choose a non-active project in `Edit Batch` (only active ones are listed)
* Could platforms be sorted alphabetically please?
* Could project be sorted first by `Enduring`, as they are now, then by project name?
* For Show/Edit batch, only active projects are listed. So, the project isn't displayed (see below)

screencast

IanMayo:
Thanks @TahaKhanAbdalli - I can't test the network calls through Heroku, partially because it lacks the high volume data, and partially because it doesn't make any network calls :-D

TahaKhanAbdalli:
> Thanks @TahaKhanAbdalli - I can't test the network calls through Heroku, partially because it lacks the high volume data, and partially because it doesn't make any network calls :-D

You can turn on the loggin from user menu and it will show the queries the dataprovider is making which at the end are api calls.

IanMayo:
Thanks @TahaKhanAbdalli  - note: I still have to provide `Vault`, `ORganisation` and `Department` for new batches.  When there is only one active value present it should be pre-selected, and the control marked as disabled/read-only.
image

IanMayo:
Taha - currently these `single-value` controls don't get assigned the default value:
image

PR #749, opened 2023-10-09
project column label changed on live and all items

### LINKED ISSUE #741: Lookup label for `project` column
Our `configData` table allows the label for `projects` to be over-ridden.

We should use the `projectName` field to provide the label for `project` in `Live Items` and `All Items` list pages.

TahaKhanAbdalli:
fixes #741

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -157,6 +157,7 @@ const getFilters = (
       source='project'
       key='project'
       label={projectLabel}
+      inputProps={{ label: projectLabel }}
       sort={sort('id')}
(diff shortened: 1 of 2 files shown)


IanMayo:
Thanks @Taha,
we also need to replace the name in the drop-downs for `Columns` and `Add filter`:
image

It's easier to test when running from SQLite backend, since that uses `Projekt`.

IanMayo:
One missing location, please @TahaKhanAbdalli .

In the `Project` filter, it is still using the `Project` label. I've renamed the config to `Projext`:
image

IanMayo (approved):
Tested and working.

PR #760, opened 2023-10-11
only showing active protection values on item add

### LINKED ISSUE #755: Only show active values for "Protection" drop-downs in AddItem
When adding a new item, we have the 3 drop-downs for itemCode, itemHandle and itemCave.

For `Add`, we should only show the `Active` entries for these drop-downs.

image

TahaKhanAbdalli:
fixes #755


IanMayo:
Thanks @TahaKhanAbdalli 
Have had a play. I was able to edit an existing item to use non-active values, but they did not render.  It _looks-like_ the lookup is only happening for valid items.

IanMayo (approved):

PR #763, opened 2023-10-11
removed VITE_API_URL from .env

### LINKED ISSUE #756: Avoid requirement for VITE_URL
The VITE_URL in `.env` lets us specify the path to the Soul backend.

Now that we are hosting RCO within Soul we don't need the backend URL to be specified, we should be able to determine it from the current URL.

Note: it won't be acceptable to have this as localhost.  We need to test it from another device, so it's a none-localhost URL.

Note2: we _may_ need another strategy for when we wish to separate the two server processes, in dev mode.

TahaKhanAbdalli:
fixes #756

### DIFF
--- a/.env
+++ b/.env
@@ -3,5 +3,4 @@ VITE_DATA_VERSION="2"
 # change this value to wipe cached UI prefs (including filters and visible columns)
 VITE_APP_VERSION="1"
 
-#API URL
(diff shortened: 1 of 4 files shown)


IanMayo:
Close. Fixed in #769

PR #779, opened 2023-10-18
reset password facility provided

### LINKED ISSUE #776: Power-User able to clear user password
We could do with a _temporary_ password reset facility.

If the current user is an `rco-power-user`, then when viewing a `User Show` page, they will have a `Clear password` button.

When this is clicked, a dialog will ask them to enter a reason for the reset (or Cancel). If `OK` is pressed then two entries are added to the audit log:  one with event id of `PWD-RESET` the ID of the person who had their password reset, and one with `RESET-PWD` for the current user.

TahaKhanAbdalli:
fixes #776


IanMayo (changes requested):
Hello @TahaKhanAbdalli - since the audit entry for "password reset for this user" has the id of the current user already, I don't think we do actually need two audit entries.

image

Specifically, I don't think we need the entry for "Password reset successfully"

IanMayo (approved):
Tested and working.

PR #797, opened 2023-10-23
only active protective markings on create

### LINKED ISSUE #787: Only show `Active` protective markings in `Add Item`
In the `item` form, for `Edit` we need to show all protective markings, but in `Add` it should just be active ones.

TahaKhanAbdalli:
fixes #787


IanMayo (approved):
Tested and working

Issue #792, opened 2023-10-20
Problem generating Batch number


IanMayo:
When users create new batches, a Batch Number like `VNaN/2023` is created.

Since I _think_ the logic looks for the last Batch Number found, once there is one corrupted name, all subsequent ones will be too.

Ian to investigate further onsite.

Also, produce `instrumented` version of RCO that includes debug lines in `BatchLifeCycle.generateBatchId()`

@TahaKhanAbdalli - I think I've found the issue here.  `batchNumber` values look like `1/2023` `2/2023`, and so on to `145/2023`.  I _think_   the generateBatchId code is failing.

Aah, the code is still expecting "V12/2023" (with a leading "V").  They no longer have the leading V.  At least, I don't think they do. To be double-checked.

OK @TahaKhanAbdalli  - have investigated.   The `batchNumber` values on their database look like `13/2023`, not `V13/2023`. Could you make the fix please?  I _guess_ we should update the mock batch generation code, too.

PR #798, opened 2023-10-23
DEFER - Introduce debug steps for batch id generation

IanMayo:
Introduce instrumentation for batch id generation, to investigate fault in more detail.

Don't bother merging this until Ian has inspected real batch reference numbers - that may offer sufficient explanation.


IanMayo:
Close. Issue solved in #792

PR #800, opened 2023-10-23
removed password field for user

### LINKED ISSUE #786: Remove `Password` field for `Edit User`
Our process for forgotten passwords is for a power user to reset the password.

So, we don't ask for a password when adding a new user, and we don't have a password control when editing a user.

@IanMayo to check if they have `active` field in their database.  If they don't remove it from our SQLite schema and mock datasets.

**We need to remove user.active from the backend and schema definitions**

TahaKhanAbdalli:
fixes #786


TahaKhanAbdalli:
Yeah, the `active` is the problem for long, I did `notice` it but I forgot to report

IanMayo (approved):
The version tested was broken, because of an inconsistency in the backend schema. (user-active needs to be removed).
But, this PR contains a valid solution for the front-end.

PR #804, opened 2023-10-24
batch number generation bug fixed

### LINKED ISSUE #792: Problem generating Batch number

TahaKhanAbdalli:
fixes #792


IanMayo (approved):
Tested and working

PR #809, opened 2023-10-24
user role made optional in user form

### LINKED ISSUE #803: Need to allow a user to have no role in `Edit User` form
Our form validation for `Edit User` requires the role to be one of `rco-user` and `rco-power-user`.  It should allow an empty field.

TahaKhanAbdalli:
fixes #803


IanMayo (approved):
Tested and working

PR #810, opened 2023-10-24
item location change bug fixed

### LINKED ISSUE #808: Cannot change item location
From `ItemList` we can select and item and change its location, providing a reason why the location is being changed (in `editRemarks`).

But, RCO is trying to update the `editRemarks` field in the relevant item.

The contents of `editRemarks` is actually intended to go into the audit log for that item.

image

image

image

TahaKhanAbdalli:
fixes #808


IanMayo (approved):
Tested and working

PR #813, opened 2023-10-24
deletion marked as unsupported

### LINKED ISSUE #801: Ensure no users can delete items
Our UI does not include delete buttons for any types of resource (things can only be made not active).

But, our permissions table doesn't explicitly deny power-users from being able to delete things.  We should explicitly state that in the data, to ensure that in future maintenance a `delete` button isn't accidentally left enabled

image

TahaKhanAbdalli:
fixes #801


IanMayo:
But @TahaKhanAbdalli - can't we implement it using the `permissions` definitions like we do for `rco-user`?

TahaKhanAbdalli:
If we do that way then we'll have to make changes to all the pages because right now the permissions of `power-user` are matched directly with `*`. If you say so, I can start working.

IanMayo:
Oh, wouldn't it just be a new version of the `rco-user` block here, but with more write permissions?

https://github.com/DeepBlueCLtd/RCO/blob/main/src/providers/authProvider/permissions.ts#L5

IanMayo:
Closing this. I'd like us to take a different, declarative approach .

PR #815, opened 2023-10-24
Trial alternate method to get IP Address

### LINKED ISSUE #814: IP address not being stored in Audit log
The audit log captures the IP address of the person responsible for the audit event.

But, valid IP addresses aren't being captured. They all appear as `127.0.0.1`

IanMayo:
Fixes #814

### DIFF
--- a/_extensions/api.js
+++ b/_extensions/api.js
@@ -5,7 +5,7 @@ const getIp = {
   method: 'GET',
   path: '/api/ip',
   handler: (req, res) => {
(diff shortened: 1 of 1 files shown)


IanMayo:
This seems to be working correctly. But, we should do a minor trial on the target network before we merge.

IanMayo:
Current strategy for IP address isn't working anyway, so we may as well have a go with this strategy.

PR #819, opened 2023-10-25
password removed from audit logs

### LINKED ISSUE #817: When user is edited, don't capture password in audit log
The audit log lists the previous values of fields for an edit.

But we have a special case.  If the password field has changed, we shouldn't log the previous value of the password.

I _Guess_ we can solve it in the relatively clunky way of checking if there is a `password` element in the delta block, and deleting it, if present.

TahaKhanAbdalli:
fixes #817


IanMayo (approved):
Tested and working.

PR #821, opened 2023-10-25
save/clone not capturing itemCode, itemCave and itemHandle fixed

### LINKED ISSUE #816: Save/Clone does not capture itemCave, itemCode, itemHandle
A user has added an item to a batch.  They set values for all the `protection` fields.

They pressed `Save/Clone`.  A new item was shown, with the previous values in the protection fields all visible.

They pressed `Save/Clone` a few more times, to create some new cloned items.

Then they went to the batch, and viewed the items. When they drilled down into the `ItemShow` page, only the `Protective Marking" was set, the others weren't.

It _looks-like_ we are correctly populating the form when cloning an existing item, but when we save that form, these new values for `itemCode`, `itemCave` and `itemHandle` are not recognised.

TahaKhanAbdalli:
fixes #816

### DIFF
--- a/src/components/ProtectionRefInput.tsx
+++ b/src/components/ProtectionRefInput.tsx
@@ -62,7 +62,9 @@ export default function ProtectionRefInput<
   const [prevValue, setPrevValue] = useState<Array<T['id']>>([])
   const {
     formState: { errors, isSubmitSuccessful },
-    setValue
+    setValue,
+    getValues,
+    reset
   } = useFormContext()
   const [valueLabel, setValueLabel] = useState<string>('')
   type SelectedIdType = T['id'] | Array<T['id']>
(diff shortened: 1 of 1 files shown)


src/components/ProtectionRefInput.tsx
```diff
@@ -164,6 +166,7 @@ export default function ProtectionRefInput<
       } else if (id) {
         createRecord(id, selectedData as number[])
       }
+      reset(getValues())
```
IanMayo, line 169: Hello @Taha - could you give me a quick description of what was broken, and how this fix works please?

IanMayo (approved):
Solution works

src/components/ProtectionRefInput.tsx
TahaKhanAbdalli: `isSubmitSuccessful` was not getting reset after first submission. So, I `reset` the form and set the values that already were there which triggered `isSubmitSuccessful` to turn to true from false.

PR #825, opened 2023-10-25
unnecessary password assigned event in audit log fixed

### LINKED ISSUE #820: Unnecessary `Password assigned` audit event
I've just changed the role for a user, from `rco-power-user` to `rco-user`.

I see this audit log for the user:
image

1. The password wasn't changed, so there shouldn't be a `password assigned` audit event
2. There is a `subjectId` field in the change record which doesn't seem valid.   It always seems to take the value `1`.  I _think_ `subjectId` is meant to be the id of the `subject` of a modification, but it wouldn't be `1` in this case.

TahaKhanAbdalli:
fixes #820

### DIFF
--- a/src/providers/dataProvider/dataprovider-utils.ts
+++ b/src/providers/dataProvider/dataprovider-utils.ts
@@ -60,7 +60,7 @@ const getActivityDetail = (
 
   const activityDetail = `{"Previous values": ${stringify(difference)}${
     editRemarks ? `, "Remarks": ${stringify(editRemarks)}` : ''
-  }, "subjectId": 1}`
+  }}`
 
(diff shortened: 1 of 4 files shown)


IanMayo:
Breaking unit test @Taha

image

IanMayo:
> @IanMayo hi there, you keep tagging me by mistake. This is NOT @TahaKhanAbdalli. Please be mindful of tagging the wrong person in the future!

Oh no, that must be frustrating - and the more it happens, the more you are offered as the default match.  I'm really sorry @taha  - I'll promise I'll try to remember in the future.

IanMayo (approved):
Tested and working

PR #836, opened 2023-10-25
filter by user fixed on audit list

### LINKED ISSUE #830: Audit Log - can't filter by `User`
The Reference / Audit log includes a filter for `User`.

This is expected to be a drop-down list of users . But, it looks like it is expecting an integer.

TahaKhanAbdalli:
fixes #830


IanMayo (approved):
Tested and working.

PR #843, opened 2023-10-26
return fixed on item list

### LINKED ISSUE #835: Cannot `Return` an item that has been `Dispatched`
In the `All Items` listing we show items that have been dispatched.

When one is selected, user should be able to select `Return` from the toolbar.  It's currently disabled.

The `Return` logic is this:
- clear `dispatchJob` field
- clear `dispatchedDAte` field
- create audit log for item, with activity type as `VAUL` (to indicate something being returned to the vault)

TahaKhanAbdalli:
fixes #835


IanMayo:
_not-quite_.  We can only return an item once the dispatch has been completed.

* setting the dispatch ID marks it being added to a dispatch
* setting the dispatchedDate indicates that it has actually been sent out.

So, we only show `Return` if there is a `dispatchDate`

IanMayo (approved):

PR #845, opened 2023-10-26
extra features for departed users

### LINKED ISSUE #818: Extra features for `departed users`
1. After a user has departed, we should disable the `password reset` button in their `user-show` form
2. After a user has deparated, they should not be able to login (retrieve user details, but reject login if their `departedDate` has passed. Show toast saying "User has departed organisation"
3. In `Users List` rename logic around `showReturn` to `showReturnItem` (to avoid confusion with following change, just in case)
## Update
4. In `Users Show`, when viewing a user that has departed, show a `Return to organisation` button in the command bar. This will set the `departedDate` to 10 years in the future.  (and remove the button from `UserList`)

TahaKhanAbdalli:
fixes #818

### DIFF
--- a/src/pages/Login.tsx
+++ b/src/pages/Login.tsx
@@ -27,8 +27,8 @@ export default function Login(): React.ReactElement {
     const staffNumber = formData.get('staffNumber') as string
     const password = formData.get('password') as string
     login({ staffNumber, password })
-      .catch(() => {
-        notify('Invalid staff-ID or password', { type: 'error' })
+      .catch((err) => {
+        notify(err.message ?? 'Invalid staff-ID or password', { type: 'error' })
       })
       .then(() => {
         const storageEvent = new StorageEvent('storage', {
--- a/src/providers/authProvider/index.ts
+++ b/src/providers/authProvider/index.ts
@@ -10,6 +10,7 @@ import { decryptData, encryptData } from '../../utils/encryption'
 import { getPermissionsByRoles } from './permissions'
 import bcrypt from 'bcryptjs'
 import { type AuditFunctionType } from '../dataProvider/dataprovider-utils'
+import { checkIfDateHasPassed } from '../../utils/helper'
 
 export const getUser = (): User | undefined => {
   const encryptedUser = getCookie(constants.TOKEN_KEY)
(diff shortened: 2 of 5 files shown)


IanMayo:
Hello @TahaKhanAbdalli - I couldn't find `User Return`, because I was looking on `UserShow` instead of `UserList`.   Users won't return in bulk, so there is no merit in having it on the list page.  The show page is a better place for it.  I've updated this in the issue (#818)

IanMayo:
The solution is quite valid @TahaKhanAbdalli  - but could we have the button down by the other  "user-operation" buttons?
image

IanMayo:
Aah, one other thing. We need to log "Returned to organisation" in the audit log.

Aah, we're trying to, but I guess there is a fault in this logic:
https://github.com/DeepBlueCLtd/RCO/blob/main/src/providers/dataProvider/resource-callbacks/UserLifeCycle.ts#L27

The logic that determines the type of event does look fragile.  Instead, couldn't we generate the audit event alongside the logic that makes the change to the user record?  I'll create a separate issue for that.

TahaKhanAbdalli:
> Aah, one other thing. We need to log "Returned to organisation" in the audit log.
> 
> Aah, we're trying to, but I guess there is a fault in this logic: https://github.com/DeepBlueCLtd/RCO/blob/main/src/providers/dataProvider/resource-callbacks/UserLifeCycle.ts#L27
> 
> The logic that determines the type of event does look fragile. Instead, couldn't we generate the audit event alongside the logic that makes the change to the user record? I'll create a separate issue for that.

This is happening because before the `departedDate` happened to be null and then updated when user departs but now our logic is changed so that departedDate is always present but sometimes in future and sometimes in past.

IanMayo:
> logic is changed so that departedDate is always present but sometimes in future and sometimes in past. I think we should have a separate issue for that.

I know, that's the _cost_ of having to work with a legacy dataset.  Some decisions made 25 years ago are impacting on our design now :-(

IanMayo:
Sorry @TahaKhanAbdalli - can we have one last change please?  Could we have a read-only version of `departureDate` in `UserShow`?  It _slightly_ explains how the backend logic is implemented.

It would have a `Departure Date` as a label, and be shown on a line of it's own just above `Depart Organisation`.

Hmm, can we make `Return to Organisation` the same height as the other two buttons?  If that means we have to lose the icon, then that's ok.

IanMayo (approved):
Tested and working.

PR #850, opened 2023-10-26
wrong item count fixed on user muster list and vault location muster list

### LINKED ISSUE #837: Wrong item count in vault muster list
For vault locations we offer a muster list.

At the bottom left of the report is a count of items. Apparently the count is always 6.  No, frequently it is zero.

No, I've just checked. The `count` comes from a list of items on loan.  That is relevant when `ReportSignature` is used for `UserMusterList` (select a User from `UserList`, and click on `User Muster List`), but not for `VaultLocationReport` (select a `Vault Location` from `Vault Locations` and click on `Vault Muster List`).

I guess it is a filter that needs to be passed in. Hmm, or we do the `count` in the parent component, and pass the value in as a prop.

Also note that after a hard refresh, the count of items at different protective markings (bottom right) is empty:
image

TahaKhanAbdalli:
fixes #837


src/components/VaultLocationReport.tsx
```diff
@@ -172,7 +172,12 @@ export default function VaultLocationReport(props: Props): ReactElement {
                       label='Muster remarks'
                     />
                   </ItemsReport>
-                  <ReportSignature id={id} />
+                  <ReportSignature>
+                    <Count
+                      resource={constants.R_ITEMS}
+                      filter={{ vaultLocation: id, loanedTo: undefined }}
```
IanMayo, line 178: Aah, well spotted.  But, for SQLite to work, I think `loanedTo` needs to be `null`.  We also need to filter for `dispatchedDate: null` and `destroyedDate: null`.

_Hopefully_ we're using `null` across the Mock and SQLite databases now.

TahaKhanAbdalli: The `filter` is just being copied from where the `items` are being listed. Changing the filter in the `Count` may result in different number then actually items listed. So changing here means changing above also

image

IanMayo: Yes - that's well spotted, thanks.

IanMayo (approved):
Tested and working

PR #856, opened 2023-10-27
Reveal Password button

### LINKED ISSUE #855: Include `Reveal` button in password box
We had a few issues with users being locked out of RCO

On one occasion this was because of the caps-lock key being on.

We should add a `reveal` button to the password box, both on the login page, and in the password change dialog.

This seems to be the solution:
https://mui.com/material-ui/react-text-field/#input-adornments

IanMayo:
Fixes #855


IanMayo:
Hello @TahaKhanAbdalli - I could do with a hand from you with this, please.

I've done the mechanics, but I'd appreciate you sorting the location of the icon - so it's actually inside the password field, please.

PR #859, opened 2023-10-27
858 pending item location

### LINKED ISSUE #858: Improve `LOCATION` messages
Sometimes we select an item, but we can't make any changes to it.

This is because it is pending destruction or dispatch.

Improve the `Location` column messages, so we see the status of pending items.

IanMayo:
Fixes #858


IanMayo:
image

PR #860, opened 2023-10-30
user audits refactored

### LINKED ISSUE #852: Refactor generation of audit entries for User Record
We currently analyse the change to the User Record to determine what type of audit entry to create.

But, the code looks fragile:
https://github.com/DeepBlueCLtd/RCO/blob/main/src/providers/dataProvider/resource-callbacks/UserLifeCycle.ts#L27

Instead, it would be better to generate the audit entry alongside the code that is changing the User Record.

TahaKhanAbdalli:
fixes #852 
_Note:_ Need to know if the activity detail is ok or should be changed? Also need to discuss `User Edit` activity type in `beforeUpdate`

### DIFF
--- a/src/hooks/useAudit.ts
+++ b/src/hooks/useAudit.ts
@@ -1,7 +1,7 @@
 import { useDataProvider } from 'react-admin'
 import { trackEvent, type AuditData } from '../utils/audit'
 
-type AuditFunction = (args: AuditData) => Promise<void>
+export type AuditFunction = (args: AuditData) => Promise<void>
 
 export default function useAudit(): AuditFunction {
   const dataProvider = useDataProvider()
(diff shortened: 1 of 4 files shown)


IanMayo:
Test failing, @TahaKhanAbdalli

TahaKhanAbdalli:
> Test failing, @TahaKhanAbdalli

Yeah, needed to discuss that. We don't need `AuditType.EDIT` now because we're auditing where the user is updating which means there will be no audit entry in `beforeUpdate` lifecycle. Is that right? If yes, I'll update the `tests`

IanMayo:
I'm pretty sure we'll still capture `AuditType.Edit`.  When a player makes an edit to anything, we still create an audit entry for it.

IanMayo (approved):

PR #861, opened 2023-10-30
edit toolbar not considering form validation fixed

### LINKED ISSUE #857: EditToolbar doesn't consider the form validation in ReferenceDataForm
Some forms are not validating.  Here is a scenario: 
* Go to Media Type . 
* Click on Create button. 
* Type something in the Name field and then remove just to make the save button enable. 
* And click on save. It will not validate the name field

`ReferenceDataFrom` is rendering `EditToolBar` component with `onClick` handler to save the form which doesn't consider if the form is valid before saving. We need to make sure the form is valid before saving.

TahaKhanAbdalli:
fixes #857


IanMayo (approved):
EditToolbar form authentication working

Issue #832, opened 2023-10-25
New dispatch numbers not generated


IanMayo:
When new dispatches are created, they mistakenly have a number like `RN/NaN/2023`

This is probably down to the receipt note generation algorithm.

There is a lot of legacy data with numbers like `VAULT/23/2023`, `VAULT/24/2023` and so on. I _Think_ this is breaking the algorithm,

Duplicate issue in data, and resolve.

Here is the pattern to generate a new dispatch number:
- retrieve last row from dispatch table
- identify current year
- get `name` from dispatch row
- generate year from last 4 characters
- if year is previous to this one, then counter is `1`
- if year is equal to this one:
    - extract the first number from the value
    - increment the number

The new `name` will be formatted as `RN/12/2023`.

PR #862, opened 2023-10-30
Make dispatch, batch, destruction IDs look like target data

IanMayo:
Supports #832 

Legacy dispatches don't look like new ones, which is breaking our logic to generate new reference numbers.

image

Our code should handle the prefix being either `VAULT` or `RN` in the dispatch reference number.

So, I suspect we need to look for the slash, first numeric value rather than a specific prefix.


IanMayo:
Push through merge, so database updates for configData can be made.

PR #866, opened 2023-10-30
dates fixed for project and item

### LINKED ISSUE #854: Dates for Projects and Items
* [x] Some projects start and finish on the same day, without a time element. So, the start and end time could be the same value.  We need to allow for this.
* [x] Some items do not have start or end dates, so it must be possible to not provide them
* [x] Also, some items may be labelled with one of the date values, with users using their knowledge of the context to decide whether to record it as a start or end date.

TahaKhanAbdalli:
fixes #854


IanMayo:
We need to allow the start date to be omitted:
image

Or for the end date to be omitted:
image

Or both dates omitted:
image

(note: we do need to allow for the dates / times to be the same.

IanMayo (approved):

PR #868, opened 2023-10-30
headers and footers added from configData

### LINKED ISSUE #848: Need headers and footers on reports
For the assorted reports we should put a protective marking in the headers and footers of each page.

The value for this should come from a new `headerMarking`  in the `ConfigData` structure (and in SQLite).

I can do the SQLite change.

(small task, but I don't think we've worked with headers/footers before. Would like it to be done for pilot phase so staff can print/issue actual reports)

While we're doing this, it would be good to include these two items please:
* [ ] Page border for reports looks like it's currently 2,5cm. Could we reduce it to 1cm please?
* [ ] Staff can print `Muster Lists` for multiple users, or for multiple vault locations.  Could the page count be per user/per vault-location (if possible). Then the document for one user/vault-location is consistent with itself.

TahaKhanAbdalli:
fixes #848


IanMayo:
* [ ] When I print the listing for 3 vault locations, there is just one overall page count (1/3).  If possible, the users would like the page count to be per vault-location. They have some vault locations with hundreds of items - so having the count per-location keeps the document manageable
* [ ] Could the header and footer have the same size font on the report (they seem the same size in print dialog, but it's distracting to have them as different sizes on the first report page we view
* [ ] Could the header and footer align centrally please?  The offset may be because of the `X` cancel button at the top-right in the print preview.  Can we hide/remove that button.

TahaKhanAbdalli:
> * [ ]  When I print the listing for 3 vault locations, there is just one overall page count (1/3).  If possible, the users would like the page count to be per vault-location. They have some vault locations with hundreds of items - so having the count per-location keeps the document manageable

Sorry, I couldn't find where the page numbers are being printed after the print dialog of the browser is opened.

IanMayo:
> Sorry, I couldn't find where the page numbers are being printed after the print dialog of the browser is opened.

Aah, they are provided by Operating System, not VAL. So, we're almost certainly not in control of them.

IanMayo:
Thanks @TahaKhanAbdalli - it's not quite right.  I'm asking for a report for vault locations 8 and 14.   I see header and footer on the first page, but only footer for following pages.

IanMayo:
One last request for this PR, please. Can we please reduce the vertical space in the protective-marking count table please?
image

IanMayo (approved):

PR #873, opened 2023-10-31
prefix added for dispatch and destruction

### LINKED ISSUE #864: Prefix for dispatch note reference
A dispatch will have a reference like "V/12/2023" (legacy), or "RN/12/2023" (new items).

This reference is present on the dispatch form.  When it occurs on the form, it should have a prefix that is taken from the `configData` object, from a field called "reportPrefix".   This will be a value like "DBC/VAL"

So, the dispatch form, hastener, and destruction certificate should show a reference like "DBC/VAL/RN/12/2023".

@IanMayo  to introduce config value, @TahaKhanAbdalli to implement displaying it.

Requirement clarified [here](https://github.com/DeepBlueCLtd/RCO/commit/b702955c125865cc445fd148ad835741b505f12e)

TahaKhanAbdalli:
fixes #864

### DIFF
--- a/src/providers/dataProvider/resource-callbacks/DispatchLifeCycle.ts
+++ b/src/providers/dataProvider/resource-callbacks/DispatchLifeCycle.ts
@@ -21,13 +21,15 @@ const lifeCycles = (
     const createdByAt = withCreatedByAt(record)
     return createdByAt
   },
-  afterCreate: async (
+  afterCreate: async function afterCreate(
     record: CreateResult<Dispatch>,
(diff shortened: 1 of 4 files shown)


IanMayo:
Sorry @TahaKhanAbdalli - the requirement wasn't clear enough.

This issue isn't about changing how reference numbers are generated.  It's just about cosmetically putting that bit of text in front of the reference when it's on destruction/dispatch paperwork.

image

image

image

IanMayo (approved):
Tested and working

PR #877, opened 2023-10-31
destruction reference number faulty logic fixed

### LINKED ISSUE #865: Faulty logic for generating new destruction number
We currently generate a new destruction number by taking the next available database id.

This doesn't work in the legacy context, where there are thousands of legacy destructions.

We need to generate the new number by inspecting the reference number for other destructions this year.

@IanMayo to inspect legacy data to verify pattern:
* do we have valid `createdAt` dates  **yes**
* do we need to retrieve the past 100, filter for those in this year, find the highest counter, and increment it?

Logic to generate destruction number:
- determine current year (2023)
- retrieve last row from destruction table (ID in descending order). extract `name` field
- the format will either be `VAULT/10/2023` or `DC/10/2023`.   Extract last 4 digits, convert to integer.  If they are for previous year, counter is 1 - since this is the first destruction of this year, If they are for current year:
    - extract first number from reference (see trick [here](https://github.com/DeepBlueCLtd/RCO/blob/main/src/providers/dataProvider/resource-callbacks/BatchLifeCycle.ts#L54) )
    - increment it 

The new reference will be formatted as "DC/2/2023`.

TahaKhanAbdalli:
fixes #865


TahaKhanAbdalli:
In order to create reference in `lifeCycle`, it needs to be nullable in the `SQLite`.

IanMayo:
> In order to create reference in `lifeCycle`, it needs to be nullable in the `SQLite`.

On it. Will fix it in this branch.

IanMayo (approved):

PR #878, opened 2023-11-01
dispatch number generation fixed

### LINKED ISSUE #832: New dispatch numbers not generated

TahaKhanAbdalli:
fixes #832

### DIFF
--- a/src/providers/dataProvider/dataprovider-utils.ts
+++ b/src/providers/dataProvider/dataprovider-utils.ts
@@ -163,33 +163,6 @@ export const extendLifeCycle = (
   ...callbacks
 })
 
-const getReferenceIdFromField = (reference: string): number => {
-  // split into 3 elements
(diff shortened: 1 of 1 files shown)


IanMayo (approved):
Tested and working

PR #880, opened 2023-11-01
reporting api call failures to user

### LINKED ISSUE #841: Report database failures in popup
Database failures are logged to the console.

But, this means RCO can silently fail, without the user being aware (MBa established this when she was unable to create a new mediaType).

We should ensure users are made aware of backend failures, via a popup.  The full error should also still be logged to the console.

Note: this may relate to the issue: https://stackoverflow.com/a/57686799/92441

TahaKhanAbdalli:
fixes #841


IanMayo:
I just tried to test this, in this way.  I modified the `item` being stored in `afterCreate` for `ItemLifeCycle`, by adding an extra `day` field:
image

I then added a new item.  The backend threw an error, which went to the console, but was not shown in the UI.

IanMayo:
I've attached a PDF that matches what I saw on the screen.

At the bottom of page 2,  `V/12/2022/34` is partially missing.  At the bottom of page 3, `V/12/2023/64` is completely missing.

[VAL.pdf](https://github.com/DeepBlueCLtd/RCO/files/13238307/VAL.pdf)

IanMayo:
Taha - this is a partially complete solution ,right?  You've just handled `items`, and we need to handle changes for the other resources too, right?

TahaKhanAbdalli:
All the things that are automatically handled by `react-admin` are fixed. e.g. creation, listing and updation. But where we ourselves are making calls through `hooks` or `dataprovider` they are embedded deep in the code and need time to `analyze` where such instances are and maybe it will be done `incrementally` because they can't be identified all at once.

IanMayo:
Error logging happening for pushing non-compliant `Item` objects, but not for `dispatch` objects.  So, it appears we have to apply the `ItemLifeCycle` change to the other life-cycles.

IanMayo:
Thanks Taha,  I was able to make this edit, which triggered a database error, which didn't get reported:
image

I know this is me deliberately breaking the app, which won't happen in the real app, but it's an example of something unexpected happening.  So, I think we need to cover it with our backend error reporting, please.

TahaKhanAbdalli:
I have fixed this but as already mentioned that there may be many such `instances` which will be updated gradually.

TahaKhanAbdalli:
I have fixed this one but there maybe many such instances that we can't `identify` at once. IMHO, these would be done gradually as we `identify` more places where we are using `try-catch` or any other way of `exception handling`.

IanMayo:
> I have fixed this one but there maybe many such instances that we can't `identify` at once. IMHO, these would be done gradually as we `identify` more places where we are using `try-catch` or any other way of `exception handling`.

I get your point, but as approach "go-live" we don't want backend interactions to silently fail.   It would be "sub-optimal" to only learn about where this needs catching in the post-event analysis of corrupted data.  The two-week pilot phase (which we are about to start) is ideal time to learn about backend troubles - and the easiest way to learn about them is for users to report these red popups.

So. once bigger bugs/issues are out of the way, I think I need you to work through all that code. Hmm, I guess there aren't more than a dozen or so types of backend interaction.

But, I'll merge this PR so that we can get the code into test.

IanMayo (approved):
Incomplete, but working.

PR #883, opened 2023-11-02
count missing from muster list fixed

### LINKED ISSUE #869: Count missing from muster lists
We have two types of muster list:
- user muster list
- vault location muster list

Beneath each listing is a count of the number of instances of each Protective Marking in that list.

There is an intermittent fault, where the count is empty.

image

After a refresh, frequently the issue goes away:
image

TahaKhanAbdalli:
fixes #869


IanMayo:
Thanks @TahaKhanAbdalli - I'll give this a go.

TahaKhanAbdalli:
> Thanks @TahaKhanAbdalli - I'll give this a go.
Still testing it. Not sure if it'll work. It should be draft.

IanMayo (approved):

PR #884, opened 2023-11-02
records missing on page-break fixed

### LINKED ISSUE #876: Some rows missing from muster list at page break
In print preview, some items are missing from the User Muster List.  I've seen the same behaviour on the Vault Location Muster list.

I did this by loaning blocks of consecutive items to myself.  At the page breaks, sometimes an item is missing.

I guess we need to insert some whitespace, somewhere.

image

image

TahaKhanAbdalli:
fixes #876


IanMayo (approved):
Working perfectly :-)

Issue #888, opened 2023-11-03
Add department to `richItems`


IanMayo:
Users may need to filter items by `department`.  So, add it to the `richItems` VIEW.

Then it can be added as a filter.

PR #890, opened 2023-11-06
add new fields to `richItem` definition

IanMayo:
Supports #888

I've done the bit to update the `types.d.ts` but we still need the data to be added to the `richItem` collection in the JS mock backend.

### DIFF
--- a/src/utils/generateData.ts
+++ b/src/utils/generateData.ts
@@ -332,6 +332,8 @@ interface Params {
   project: Project[]
   platform: Platform[]
   item: Item[]
+  department: Department[]
+  vault: Vault[]
 }
 
 export const generateRichItems = async (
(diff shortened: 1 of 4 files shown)


src/utils/generateData.ts
```diff
@@ -366,12 +368,30 @@ export const generateRichItems = async (
       constants.R_PLATFORMS,
       params
     )
+    const { data: fetchedDepartments } = await dataProvider.getList<Department>(
+      constants.R_DEPARTMENT,
+      params
+    )
+
```
IanMayo, line 375: this next block should refer to vaults @TahaKhanAbdalli

src/utils/generateData.ts
```diff
@@ -386,15 +406,21 @@ export const generateRichItems = async (
...
+  departments: Department[],
+  vaults: Vault[]
 ): RichItem[] => {
   const richItems: RichItem[] = []
   for (const item of items) {
     richItems.push({
       ...item,
       id: item.id,
       platform: generateRandomNumber(0, platforms.length),
-      project: generateRandomNumber(0, projects.length)
```
IanMayo, line 397: note: the vaults and departments tables don't have an integer id - it's a string that's a short version of the actual name.  So, you need to get the row at a random number index, and store the `id` for that row.

IanMayo: Oh, hold on.  I see you've tried an alternate strategy for this, but it looks a bit fragile (including if we change what department ids look like, or introduce a third vault.

I'd rather you get a random row, and use the id for that row, please.

IanMayo:
Unable to verify from current machine. Will test later.

PR #893, opened 2023-11-07
prefix added for item number and batch number

### LINKED ISSUE #886: Don't use "V" prefix, use first letter of "vault" attribute
There are a number of places where we cosmetically prepend "V" to batch and item numbers.

This doesn't correctly support "Legacy" items.

I have added "vault" to the "richItem" view.  So, when we display batch or item numbers, we should prepend the first letter of the `vault` field (which is the `id` value of the `Vault` table).

So, we should see `V` or `L` in the batch/item number (`V12/2023` and `V12/2023/4` or `L12/2023` and `L12/2023/45`).

TahaKhanAbdalli:
fixes #886

### DIFF
--- a/src/components/TopToolbarField.tsx
+++ b/src/components/TopToolbarField.tsx
@@ -3,7 +3,7 @@ import { Show, type ShowProps, TextField } from 'react-admin'
 import { type ReactElement } from 'react'
 import { type SystemStyleObject } from '@mui/system'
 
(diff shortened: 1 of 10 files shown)


IanMayo:
Hello @TahaKhanAbdalli  - I've found another place where we need the prefix:
image

Aah, and another (if the audit relates to an `item`):
image

And this one:
image

Fixing the previous one will probably fix this one, too:
image

PR #899, opened 2023-11-08
active filter added on user listing

### LINKED ISSUE #788: Users Table - introduce filter for `Active`
There are lots of users in the database, and lots of legacy users, for whom the departed date has passed.

We should introduce a filter for `Active`. This actually filters for users with a departed date in the future.

This filter should be `On` by default.

Aah, Taha said the `departedDate` in the `user` table wasn't in ISO format.  @IanMayo to check mock backend data. Note: for the data on the target network, dates all seem to be in this format: `2009-05-02 12:00:00`. Filtering by `DateRange` on the `Live Items` page seems to work correctly.

Hmm, I _think_ that when VAL generates (or updates) this data, it produces ISO dates.  Is it possible to write content to support both date formats?  Yes, I've just marked a user as departed, and it has a value like this: `2023-11-07T10:23:24.454Z`

TahaKhanAbdalli:
fixes #788


IanMayo:
@TahaKhanAbdalli - is it possible for the `Active` filter to be applied by default please?

IanMayo (approved):
Tested and working.

PR #900, opened 2023-11-08
prefix changed for legacy data support

### LINKED ISSUE #896: Compatibility with legacy data
For legacy destructions, the stored value of `name` is formatted as `VAULT/2/2023` (including the valid vault number).

When we create new destructions we should generate the `name` to match the above.

The users want the prefix to always be the string `VAULT` - and be sensitive to a `vault` drop-down list.  So, our current `reference generation` code remains valid, but we provide the `VAULT` prefix to the function rather than DC or RN.

It's the same for dispatches.

So, we should cosmetically change the Destruction Form and Dispatch forms to include the a reference like this "DBC/RCO/VAULT/4/2023` (without `DC` or `RN`).  This cosmetic change does not need to be in the title of DispatchShow or DestructionShow.

TahaKhanAbdalli:
fixes #896


IanMayo (approved):
Tested and working

PR #904, opened 2023-11-08
welcome page items on loan approach changed

### LINKED ISSUE #812: Fix to list of users with items on loan
On the welcome page we show a list of users with items on loan.

To do this, we pull a list of all items on loan from the backend, and collate a list of unique user ids.

While this effectively still happens in an instant, it involves a lot of data (the full details for 500 items are pulled to the client), and _could_ bring unnecessary loading on the backend.

Instead, we could define a SQL view that contains a list of users with items on loan. We could even count the items on loan.

Then the `On loan` listing on the home page would just be a summary of this view.  There could be 20+ users with items on loan.  So, allow list to take full right-hand column of welcome page.  We should also present this list in ascending alphabetical order.

Here is the SQL:
```
CREATE VIEW loanItems AS
SELECT
  COUNT(i.id) as numItems,
  i.loanedTo as userId,
  u.name
from 
  item i
INNER JOIN "user" u ON
  i.loanedTo = u.id
GROUP BY
  u.name,
  i.loanedTo
```

We can call the view "loanUsers"

Aah, we have to `mock` the `VIEW` in JS.  Let's just run through the data once, on page load.  If in the mock JS there aren't any items on loan, we should loan a few.

TahaKhanAbdalli:
fixes #812

### DIFF
--- a/src/components/Recent.tsx
+++ b/src/components/Recent.tsx
@@ -62,7 +62,7 @@ interface Props<T> extends DatagridProps {
   rowStyle?: (data: T) => SxProps
 }
 
-function Column<T extends Batch | User | Dispatch>(
+function Column<T extends Batch | Dispatch | LoanUser>(
   props: Field<T>
 ): React.ReactElement {
   const { source, reference, component, label = '' } = props
@@ -127,7 +127,7 @@ export function RecentCard(props: RecentCardProps): React.ReactElement {
     </Box>
   )
 }
-export default function Recent<T extends Batch | User | Dispatch>(
+export default function Recent<T extends Batch | Dispatch | LoanUser>(
   props: Props<T>
 ): React.ReactElement {
   const {
--- a/src/components/SourceField.tsx
+++ b/src/components/SourceField.tsx
@@ -28,6 +28,7 @@ const SourceField = <
     | Project
     | ProtectiveMarking
     | MediaType
+    | LoanUser
 >(
   props: SourceFieldProps<T>
 ): React.ReactElement => {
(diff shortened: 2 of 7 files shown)


IanMayo:
@TahaKhanAbdalli - when I click on the `Items on loan` table header, it gives me `URL not found`.

IanMayo:
Note: we don't need user-name and staffNumber - since the StaffNumber is essentially the user's surname anyway on this system.  So, let's just have the user-name please - it's a little easier to read:
image

IanMayo:
Hello Taha - the previous logic was for `items on loan` to open RichItems filtered to those on loan to these individuals.

But, tell you what - that doesn't really help, since users will just click on the name of the person who is returning the items.

Let's remove the link from the title :-)

IanMayo:
Aah, I just broke it. I clicked on `StaffNumber` (to sort them by name order), and it crashed the UI:
image

TahaKhanAbdalli:
> Aah, I just broke it. I clicked on `StaffNumber` (to sort them by name order), and it crashed the UI: image
> 
> Actually, clicking on `sort` for a column in any of the 3 tables causes the error.

I noticed this error but forgot to report. The issue is that clicking on any of the tables' column header causes this crash because when the header is clicked it changes the url on which react-admin triggers a filter on all of the lists that are currently showing. So they have different fields which causes the crash. But I don't know how to handle this.

IanMayo:
> But I don't know how to handle this.

Can we disable sort on column?

TahaKhanAbdalli:
> > But I don't know how to handle this.
> 
> Can we disable sort on column?

Yes

IanMayo:
Let's do that, then please.

PR #912, opened 2023-11-10
default filters added on dispatch and destructions

### LINKED ISSUE #903: For destructions and dispatches, only show pending ones by default
A user had trouble closing an old destruction.  He didn't realise it had already been destroyed.  

The `destructions` and `dispatches` pages should have the `not dispatched` and `pending` filters applied by default.

TahaKhanAbdalli:
fixes #903


src/resources/destruction/DestructionList.tsx
```diff
@@ -43,7 +43,13 @@ export default function DestructionList(): React.ReactElement {
     setOpen(open)
   }
   return (
-    <List filters={filters}>
+    <List
+      filters={filters}
+      filterDefaultValues={
+        process.env.MOCK
```
IanMayo, line 49: shouldn't the mock data use a filter that is `false`?

src/resources/dispatch/DispatchList.tsx
```diff
@@ -78,7 +78,15 @@ export default function DispatchList(props: DatagridProps): React.ReactElement {
     : 'simple-dispatch-list'
 
   return (
-    <List hasCreate storeKey={storeKey} filters={filters}>
+    <List
+      hasCreate
+      storeKey={storeKey}
+      filters={filters}
+      filterDefaultValues={
+        process.env.MOCK
```
IanMayo, line 86: yeah, if we want to see items that have not been dispatched, we should be checking for `false`

IanMayo (approved):

PR #913, opened 2023-11-10
departed date for new users set to 10 years in future

### LINKED ISSUE #909: New users are `inactive`
When we create a new user, their departure date is empty, meaning they are effectively inactive.

For "create user" we should set a departure data that is 10 years in the future, please.

TahaKhanAbdalli:
fixes #909


IanMayo:
The implementation here works fine.  But, I've spotted a related issue.  When I mark a user as `departed`, the UI doesn't update - so it invites me to depart the user again:
screencast

IanMayo (approved):

PR #915, opened 2023-11-10
receipt note received audit entry added

### LINKED ISSUE #908: Need audit events for Receipt Note Received
We need audit log entries for these events:
* receipt note received (when clicked from the `DispatchList` page)

TahaKhanAbdalli:
fixes #908


src/resources/dispatch/DispatchList.tsx
```diff
@@ -50,6 +53,17 @@ const BulkActions = (): React.ReactElement => {
         receiptReceived: nowDate()
       }
     })
+
+    await audit({
```
IanMayo, line 57: To let users navigate to the dispatch, we need multiple entries, one for each `selectedId` please.  The `id` will go in the `dataIt` field.

IanMayo:
Sorry Taha - there's one last change on here, please.   We need some slighlty more complex logic for when we show `Receipt Note Received`:
- it should only be displayed or enabled for items that have a non-null `dispatched` at date
- it should only be displayed or enabled for items that have null `receipt Note Received` date

So, users only click it for dispatched items that have not already received the receipt note, please.

IanMayo (approved):

PR #916, opened 2023-11-14
no user can delete any resource

### LINKED ISSUE #801: Ensure no users can delete items

TahaKhanAbdalli:
fixes #801

### DIFF
--- a/src/providers/authProvider/permissions.ts
+++ b/src/providers/authProvider/permissions.ts
@@ -1,5 +1,4 @@
 import * as constants from '../../constants'
-type UserRoles = UserRole
 
 const basePermissions = {
   [constants.R_PROJECTS]: { read: true, write: true, delete: false },
@@ -15,7 +14,7 @@ const basePermissions = {
   'welcome-page': { read: true }
 }
 
-const permissions: Record<UserRoles, ResourcePermissions> = {
+const permissions: Record<UserRole, ResourcePermissions> = {
   'rco-user': { ...basePermissions },
   'rco-power-user': {
     ...basePermissions,
@@ -27,7 +26,7 @@ const permissions: Record<UserRoles, ResourcePermissions> = {
 
 export default permissions
 
-export const getPermissionsByRoles = (role: UserRoles): ResourcePermissions => {
+export const getPermissionsByRoles = (role: UserRole): ResourcePermissions => {
   const userPermissions = permissions[role]
   return userPermissions
 }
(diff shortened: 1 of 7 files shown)


IanMayo:
Thanks for this @TahaKhanAbdalli , can you remind me how we implement permissions for users without a role?  (so, not `rco-user` or `rco-power-user`).

src/providers/authProvider/permissions.ts
```diff
@@ -1,22 +1,27 @@
 import * as constants from '../../constants'
 type UserRoles = UserRole
```
IanMayo, line 2: is there a reason why we are renaming the `UserRole` type here @TahaKhanAbdalli?

IanMayo (changes requested):
@TahaKhanAbdalli  - I'm in as `power-user` but I don't have `edit` permission for the `Users` table.

TahaKhanAbdalli:
> Thanks for this @TahaKhanAbdalli , can you remind me how we implement permissions for users without a role? (so, not `rco-user` or `rco-power-user`).

I think we removed that because of the requirements.

TahaKhanAbdalli:
> @TahaKhanAbdalli - I'm in as `power-user` but I don't have `edit` permission for the `Users` table.
> 
> image

simple `rco-user` also has write permission to users. Do I remove that too?

IanMayo:
> > Thanks for this @TahaKhanAbdalli , can you remind me how we implement permissions for users without a role? (so, not `rco-user` or `rco-power-user`).
> 
> I think we removed that because of the requirements change.

Aah yes, we did that here:
https://github.com/DeepBlueCLtd/RCO/commit/eaea7fe97ed2180eb4ccd1feafc1c61ad68a49d0#diff-fbd3282a67d2fc43a6f851ad4c86884a61d677bc927eabb2abfe551148ee760e

IanMayo:
> > @TahaKhanAbdalli - I'm in as `power-user` but I don't have `edit` permission for the `Users` table.
> > image
> 
> simple `rco-user` also has write permission to users. Do I remove that too?

No, the issue is that my role was `missing` the ability to edit a user.  I'm ok with plain `rco-user` having write permission.

IanMayo (approved):

Issue #735, opened 2023-10-03
Password/authentication requirements


IanMayo:
Here is a JS password strength validator: https://github.com/tarunbatra/password-validator

This task also includes the SME review of the auth strategy.  It is acknowledged that this is a temporary strategy, to be superceded by #766 

The application will require a password policy:
* Min length: 10 characters `(lib - 10-min)`
* Min Password Age: 1 day `(5-hr)`
* Max Password Age: 120 days `(2-hr if min password age is done)`
* Password history: 5 `(6-hr)`
* Lockout Attempts: 5 `(8-hr)`
* Password must include (at least one of each): Uppercase, Lowercase, Numeric and Special Character. `(lib - 20-min)`

**Note:** - it is inferred in the above that passwords expire after 120 days, so we'll need to check that at login.

Only unlocked by an admin in this application.

* Passwords cannot include username, first name, last name, real names. `(lib - 1-hr)`
* Passwords cannot include consecutive characters (4). `(lib - 1-hr)`
* Passwords cannot match patterns; such as qwerty123 or 09876zxcvb. `(lib-1hr)`
* ~~Passwords cannot be similar to your previous five. `(3-hr if password history is done)`~~ (requirement relaxed, 16/11/23)
* Passwords stored must be salted then hashed using standards such as SHA-256 (minimum). `(already done, will verify - 30-min)`
* Cookies/tokens must be encrypted using similar standards. `(already done, will verify - 1-hr)`
* Sessions/cookies must be destroyed/expired during logout or closing the browser. `(lib - 2-hr)`

PR #919, opened 2023-11-17
password insertion extension created

TahaKhanAbdalli:
supports #735


_devExtensions/api.js
```diff
@@ -6,6 +12,87 @@ const getIp = {
   }
 }
 
+const removeOldestPassword = (db, userId) => {
+  const query = `DELETE FROM ${tableName} WHERE userId = ? ORDER BY createdAt LIMIT 1`
```
IanMayo, line 16: @TahaKhanAbdalli - can we structure the query to delete all except the most recent 5, just in case?

_devExtensions/api.js
```diff
@@ -6,6 +12,87 @@ const getIp = {
...
+  db.prepare(query).run(userId)
+}
+
```
IanMayo, line 19: I think it should be `count` rather than `cound` in the next line.

IanMayo (approved):

PR #923, opened 2023-11-23
added salt in the password

### LINKED ISSUE #920: Salt password with user-id
I had to login as someone else, who hadn't had a password set.

So, I went into the `user` table and copied my password to theirs.  I was then able to login using my password.

The down-side to this is that _theoretically_ if someone saw the user table, and saw two identical passwords they could get the password off one, and log in as another.

But, if we salt the password with their user-id (the immutable `id` field, I guess) then identical passwords will still be different for different users.

TahaKhanAbdalli:
fixes #920 

Set this password for staffNumber `20300` to `login`:
`$2a$12$qkFbUUmb5kpFMXkqC8fruedkZLo2LD6joCgVRHMrVJVyrZWriicDe`


IanMayo:
@TahaKhanAbdalli  - we don't need this PR any more, do we?

PR #929, opened 2023-11-24
renamed staffNumber to Username

### LINKED ISSUE #926: Cosmetically renamed `staffNumber` to `username`
Staff onsite have both a staff number and a username.

When adding a user, the  VLA user entered their staff-number into the box titled `Staff Number`.  That's understandable :-)

So, we don't need a database change, but ensure that everywhere that `Staff Number` appears, we show `User name` instead.

TahaKhanAbdalli:
fixes #926


IanMayo (approved):
Tested and working.

PR #938, opened 2023-12-05
changed better sqlite version

### LINKED ISSUE #937: SQLite confusion
There appears to be an issue in our login/auth processing.

The `user` data accessed by the `_extension` seems out of date.  I user had exceeded the number of login failures, and was locked out `lockoutAttempts=5`.

I cleared that using `Reset Password`.  I used the REST API to verify the lockout attempts was cleared.

When I tried to log in again, I was still locked out (password attempts).

I used sqlite-cli to view the data, and it still said `lockoutAttempts:5`.  But when I exited SQLite3 and re-entered, it had an updated value of zero.

I think we can (kind of) reproduce the issue using these steps:
1. install sqlite3. (`brew install sqlite3`)
2. navigate to db folder
3. open database `sqlite3 RCO2.sqlite`
4. view user 10 `select * from user where id=10;`
5. verify name is `Henrietta`
6. exit sqlite3 (`.quit`)
7. rename user to `Henrietta B` in VAL
8. view her data again via command line - see name is still `Henrietta`
9. exit sqlite (`.quit`)
10. view SQLite3 again - should see her name has updated.

It seems that _something-like_ this is happening: 
1. when sqlite3 opens, it doesn't handle `.WAL` changes 
2. when slqite3 closes, it updates to process WAL changes
3. when sqlite3 re-opens it contains updated data

I'm grateful for any help you can offer in moving this forward...

Doing some googling. _Maybe_ we're using multiple SQLite drivers:
https://stackoverflow.com/questions/68572090/why-data-is-lost-in-sqlite-database-with-wal-mode-on-when-connection-is-not-clos

I have checked our package.json and that for `soul`.  We're on two different versions of `Better-SQLite`. It may be worth pushing VAL back to the same as `Soul`.

TahaKhanAbdalli:
fixes #937

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@vitejs/plugin-react": "^3.1.0",
     "axios": "^1.4.0",
     "bcryptjs": "^2.4.3",
-    "better-sqlite3": "^9.1.1",
+    "better-sqlite3": "^8.1.0",
     "cross-env": "^7.0.3",
     "crypto-js": "^4.1.1",
     "dayjs": "^1.11.7",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3011,14 +3011,6 @@ better-sqlite3@^8.1.0:
     bindings "^1.5.0"
     prebuild-install "^7.1.0"
 
-better-sqlite3@^9.1.1:
-  version "9.1.1"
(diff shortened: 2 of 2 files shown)


IanMayo (approved):
Change looks logical. Will need to test on target network.

PR #940, opened 2023-12-08
Adopt optimistic record locking

### LINKED ISSUE #934: Login issues
It appears we have two login issues:

1. Reinstated user. A user role has been reinstated, using `RESET PASSWORD`.  When he tries to login using StaffNumber in both fields he gets a message: `Your account is locked. Please contact your administrator`  This is triggered by `lockoutAttempts` being greater than 5. But, I can see via the REST API that it is zero,  (this was on the client network)
2. Another reinstated user.  A user role has been reinstated using `RESET PASSWORD`.  When I try to log into that role using the staff number in both fields I get `Error Code: 400, Message: Invalid password`.  Code-inspection suggests that has failed the brypt hash comparison. But, as this user has an empty password I don't think we should be doing bcrypt comparison. (this was in a PR on heroku)

Note: I've made a new build that includes the number of lockout attempts (#933).  It also uses different types of brackets to clarify if it's the front-end or backend that is raising the error. There is no harm in this functionality remaining in VAL.

TahaKhanAbdalli:
fixes #934

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -80,6 +80,7 @@ export type ResourceTypes =
   | typeof R_RICH_ITEMS
   | typeof R_LOAN_USERS
 
+
 // preferences
 export const LOGGING_ENABLED = 'LOGGING_ENABLED'
 export const AUTH_STATE_CHANGED = 'AUTH_STATE_CHANGED'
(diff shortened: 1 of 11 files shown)


TahaKhanAbdalli:
`Pessimistic` or `Optimistic`?
 Because `Optimistic` not hide the `undoable`.

IanMayo (approved):

PR #942, opened 2023-12-13
941 editing issues

### LINKED ISSUE #941: Bug - editing issue causes catCave to be duplicated
A user had an item open with the right-hand protection field having a single lozenge/value assigned.

They edited the item, to change the finish date.   They pressed save, and gave a reason.

The right-hand protection field now had two instances of the single lozenge.  Each edit introduced another instance of the lozenge.

This _may_ be related to the logic for save/clone and save/new.  But, since this is plain `edit` we shouldn't be messing with these `catXXX` fields.

(onsite now so can't reproduce it exactly)

TahaKhanAbdalli:
fixes #941


IanMayo (approved):
tested and working

PR #945, opened 2023-12-18
Audit log on edit to user table resolved

### LINKED ISSUE #944: Need audit logs on `Edit` to `User` table
We've just changed the role for a user. 

The audit log doesn't record this event.

We already log `password reset`.  We should log other `User` edits too.  They should be flagged as `Security-Related`.

Note: for the `Audit Log`,  it would be better if we cosmetically presented a `1` in the `Security Related` column as a font-awesome tick-icon.

TahaKhanAbdalli:
fixes #944

### DIFF
--- a/src/resources/users/index.tsx
+++ b/src/resources/users/index.tsx
@@ -2,6 +2,8 @@ import React from 'react'
 import { Create, Edit } from 'react-admin'
 import UserForm from './UserForm'
 import * as constants from '../../constants'
+import useAudit from '../../hooks/useAudit'
+import { AuditType } from '../../utils/activity-types'
 
 const UserList = React.lazy(async () => await import('./UserList'))
 const UserShow = React.lazy(async () => await import('./UserShow'))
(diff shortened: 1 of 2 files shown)


src/resources/users/index.tsx
```diff
@@ -18,9 +20,23 @@ const UserCreate = (): React.ReactElement => {
 
 const UserEdit = (): React.ReactElement => {
   const path: string = `/${constants.R_USERS}`
+  const audit = useAudit()
 
   return (
     <Edit
+      mutationOptions={{
```
IanMayo, line 27: Thanks @TahaKhanAbdalli - but I could with detail of what got edited.  We can't show the password, but it would be useful to know if the role changed, username got edited, etc.   Since there are so few editable properties on the form, I guess you don't have to use the complex processing like we do in `Edit Item` - maybe just do a manual comparison, and provide a suitable description for the old values.

IanMayo (approved):
Tested and working.

PR #946, opened 2023-12-20
inactiver user token expire after 1 hour

TahaKhanAbdalli:


IanMayo:
The logic looks good here @TahaKhanAbdalli .

But, the token will still be set to expire automatically after 1 hour since logging in, won't it?  (the current implementation)

IMHO we could set the expiry date to one day - since the users do logout every evening.

TahaKhanAbdalli:
Got it. I am updating the token expiry from one hour to one day.

IanMayo:
I need to test this. Will try to tomorrow.  I'll temporarily shorten the idle time.

IanMayo (approved):
Approved via code inspection.

PR #969, opened 2023-12-22
handling subject field in auditlist resolved

### LINKED ISSUE #950: Audit List - handling of `Subject` field
We offer `Subject` as a filter in the Audit List.

It shows a series of user names.  But, the page is actually filtering on the subject-id - as an integer.

I _guess_ we should accept an integer 

I _also_ guess we should provide a `data` filter - that filters by the data-id.

These are pretty complex filters.  Maybe we should add ` (expert users only)` to their entries in the drop-down list.

TahaKhanAbdalli:
fixes #950

### DIFF
--- a/src/resources/audit/AuditList.tsx
+++ b/src/resources/audit/AuditList.tsx
@@ -15,7 +15,8 @@ import {
   ExportButton,
   useGetList,
   Link,
-  BooleanField
+  BooleanField,
+  NumberInput
 } from 'react-admin'
 import * as constants from '../../constants'
 import ActivityTypes from '../../utils/activity-types'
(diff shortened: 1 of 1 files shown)


IanMayo:
Two things to tidy, please @TahaKhanAbdalli 

1.  `Subject` and `data` aren't all that clear now that we have different resource types. Could we please rename `data` to `Subject` and `Subject` to `Object` (that's the correct interpretation of `subject` and `object` in this context.
2. Could we put the new `Subject` and `Object` filters at the end of the list of available filters please?

IanMayo (approved):

PR #974, opened 2024-01-02
966 - unique value for new mediaType id

### LINKED ISSUE #966: Can't create new MediaType
I can create a new media type on Heroku

But I can't on the target network.  

I am diagnosing what is happening via the `Network` tab.

RCO does a `POST` with the new media type.  The payload includes a new `id`.  On the target network, this is id `938`.  Note: `938` is already in use, and the media-types actually go up to `1061`.  I've tried it several times, and it is always `938`.

The `post` fails, with an error `UNIQUE constraint failed: mediaType.id`.

When I monitor what happens on Heroku, it appears to provide a valid id (an incremented version of the current max id)

I've inspected the schema of the SQLite database on the target network.  The `id` for `mediaType` is defined as `INTEGER NON NULL UNIQUE`.  I've just inspected the SQLite database for our mock data.  I see that it has the same `UNIQUE` property - so that's not the issue.

Note: the [schema definition](https://github.com/DeepBlueCLtd/RCO/blob/main/db/Scripts/full.sql#L87) says it should not be unique, but that doesn't match the actual database.

**Note:** this is another instance of RCO silently failing.  Aah, but we can't test it on Heroku or our dev machines.

IanMayo:
Fixes #966

### DIFF
--- a/src/hooks/useCustomId.ts
+++ b/src/hooks/useCustomId.ts
@@ -14,9 +14,9 @@ export default function useCustomid(): UseCustomId {
   const { getValues } = useFormContext()
   const [create] = useCreate() as any
   const resource = useResourceContext()
-  const results = useGetList(resource, {
-    sort: { field: 'id', order: 'DESC' },
-    pagination: { page: 1, perPage: 1 }
+  const {data : existingRcordes} = useGetList(resource, {
+    sort: {field: 'id', order: "DESC"},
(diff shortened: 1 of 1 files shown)


IanMayo:
Could you check this please @TahaKhanAbdalli - feel free to tidy the code, if necessary.

IanMayo:
Great, thanks @TahaKhanAbdalli

PR #975, opened 2024-01-03
resource filter in audit list issue resolved

### LINKED ISSUE #951: Audit List - resource filter
The resource filter is a free text box.

But, we actually know the set of available resources. So, we should provide a drop-down list of them in the `Resource` filter.

TahaKhanAbdalli:
fixes #951


src/resources/audit/AuditList.tsx
```diff
@@ -43,6 +43,21 @@ const SecurityRelatedFilter = ({
 
   return <Chip sx={{ marginBottom: 1 }} label={label} />
 }
+const availableResources = [
```
IanMayo, line 46: @TahaKhanAbdalli - don't we have constants for these string values?

src/resources/audit/AuditList.tsx
```diff
@@ -43,6 +43,21 @@ const SecurityRelatedFilter = ({
...
+  'user',
+  'item',
+  'batch',
+  'destruction',
+  'vaultLocation',
+  'dispatch',
+  'project',
```
IanMayo, line 53: Aah, we cosmetically rename `project` using the config data.

I think this requires this change:
- array contains pairs of elements - the resource name and the cosmetically displayed label.  For this, we'll use the config value for `project` and present `vaultLocation` as `vault location` (same with mediaType).
- when we render values for the `resource` column, we should display this cosmetic label, not the raw `resource` value.

TahaKhanAbdalli: Working on it.

src/resources/audit/config.tsx
```diff
@@ -0,0 +1,24 @@
...
+  constants.R_PLATFORMS
+]
+export const cosmeticLabels = {
+  [constants.R_USERS]: 'User',
+  [constants.R_ITEMS]: 'Item',
+  [constants.R_BATCHES]: 'Batch',
+  [constants.R_DESTRUCTION]: 'Destruction',
+  [constants.R_VAULT_LOCATION]: 'Vault Location',
+  [constants.R_DISPATCH]: 'Dispatch',
+  [constants.R_PROJECTS]: 'Project',
```
IanMayo, line 21: the value of Project needs to come from the config-file, please @TahaKhanAbdalli

TahaKhanAbdalli: got it.
I was thinking to add config based on audit and other files but never-mind.  let me add them to the constants file.

IanMayo: But note: the resource used in the filter needs to be `project` (since that's the resource/table name), but the displayed label will come from the `ConfigData` (it's `Projekt` in our mock dataset)

IanMayo:
Just one more bit - taking the `project` label from the `ConfigData`:
image

src/resources/audit/config.tsx
TahaKhanAbdalli: Should I turn `project` to `Projekt` into the resource filter?

src/resources/audit/AuditList.tsx
```diff
@@ -251,7 +284,18 @@ export default function AuditList({
           label='Security Related'
           looseValue
         />
-        <TextField<Audit> source='resource' label='Resource' />
+        <FunctionField<Audit>
+          source='resource'
+          label='Resource'
+          render={(record) => (
+            <span>
```
IanMayo, line 291: @TahaKhanAbdalli  - have a linter warning on the next line.
image

Is it worth refactoring it into an arrow function, where we can be a bit more verbose?

TahaKhanAbdalli: Let me check.

IanMayo (approved):
Tested and working

PR #982, opened 2024-01-09
live items asset report fails with project filter issue resolved

### LINKED ISSUE #957: Live Items - Asset Report fails with `Projekt` filter
In the `Live items` page, a user added a `Projekt` filter, and a project was selected.  A filtered set of items were shown.

When he clicked on `Asset Report` an error was shown `No such column: item.project`

I _guess_ it is because the report is looking at the `items` table, not the `richItems` one that introduces `projeckt` and `platform`.

TahaKhanAbdalli:
fixes #957

### DIFF
--- a/src/resources/items/ItemsReport.tsx
+++ b/src/resources/items/ItemsReport.tsx
@@ -99,7 +99,7 @@ export default function ItemsReport(
   const { footer, children, headStyle = {}, sx = {}, ...rest } = props
   return (
     <List
-      resource={constants.R_ITEMS}
+      resource={constants.R_RICH_ITEMS}
       pagination={false}
       actions={false}
       sx={{ margin: '20px 0' }}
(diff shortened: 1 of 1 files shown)


src/resources/items/ItemsReport.tsx
```diff
@@ -165,6 +165,18 @@ const referenceFields: RefFieldType[] = [
     resource: constants.R_BATCHES,
     source: 'batchNumber',
     label: 'Batch'
+  },
+  {
+    name: 'project',
+    resource: constants.R_PROJECTS,
+    source: 'name',
```
IanMayo, line 172: the label for `Project` should come from the config data, like here:
`ConfigData?.projectName`

IanMayo:
Software working, but one minor issue.  If the user has provided a filter for `Project`, we need to cosmetically change the label to the one from `configData`:
image

So, in our sample data it will show `Projekt:` rather than `Project:`.

IanMayo (approved):
Tested and working

PR #985, opened 2024-01-15
Change user role triggered reset password issue resolved

### LINKED ISSUE #979: Changing user role triggered "Change password"
A user changed their role (from Power User to ordinary user, I think).

They then logged out, and closed the browser.

When they returned to RCO, they were at a "Enter password" dialog for their role.

This suggests that "change role" had cleared the password, so triggering a "Enter password" - but that's just a guess, it may be another issue.

But, it also suggests that the user cookie is stored even when the browser tab is closed - since that must be how RCO knew which role to "Enter Password" for.  When the browser has closed, and a new RCO tab opens, RCO should not have a cookie stored, and the user should have to enter their user-name to get in.

TahaKhanAbdalli:
fixes #979

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,7 +16,7 @@ import {
   VisibilityOff
 } from '@mui/icons-material'
 import { getDataProvider } from './providers/dataProvider'
-import rcoAuthProvider from './providers/authProvider'
+import rcoAuthProvider, { removeToken } from './providers/authProvider'
 import { useForm } from 'react-hook-form'
 import * as yup from 'yup'
 
@@ -232,6 +232,12 @@ function App(): React.ReactElement {
   }
 
   useEffect(() => {
+    const storedSessionData = sessionStorage.getItem('login')
+    if (storedSessionData !== null) {
+      console.log('Stored Value:', storedSessionData)
+    } else {
+      removeToken()
+    }
     const storedValue = localStorage.getItem(constants.LOGGING_ENABLED)
     if (storedValue !== null) {
       setLoggingPref(storedValue === 'true')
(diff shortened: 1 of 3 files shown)


IanMayo:
Thanks @TahaKhanAbdalli - could you through the strategy for this fixes the issue please?

TahaKhanAbdalli:
While editing roles, we mistakenly triggered a password reset by including an empty string in the PUT request payload. To fix this, I removed the empty string password from the payload, avoiding unintended password resets. 
Additionally, I implemented session storage to enable users to log in again if they close the browser directly. Session storage stores data temporarily during a user's session, but it gets cleared when the user closes the browser tab or window, ensuring privacy and security.

src/App.tsx
```diff
@@ -232,6 +232,12 @@ function App(): React.ReactElement {
   }
 
   useEffect(() => {
+    const storedSessionData = sessionStorage.getItem('login')
+    if (storedSessionData !== null) {
+      console.log('Stored Value:', storedSessionData)
```
IanMayo, line 237: I don't think we need to log this message, once we know the code is working.

IanMayo:
Could you add a comment near where you set the token, to explain what is happening please?

src/App.tsx
```diff
@@ -232,6 +232,12 @@ function App(): React.ReactElement {
   }
 
   useEffect(() => {
+    const storedSessionData = sessionStorage.getItem('login')
```
IanMayo, line 235: we should use a constant instead of `login`, please.  Where we define the constant, we should explain what it is being used for.

src/App.tsx
```diff
@@ -232,6 +232,12 @@ function App(): React.ReactElement {
...
+    if (storedSessionData !== null) {
+      console.log('Stored Value:', storedSessionData)
+    } else {
+      removeToken()
```
IanMayo, line 239: @TahaKhanAbdalli - could you please rename `removeToken()` to `removeUserToken()` - since it is removing a specific token.

TahaKhanAbdalli: Sure.

IanMayo (approved):

PR #988, opened 2024-01-18
rco-user not able to edit any user profile

### LINKED ISSUE #978: Ordinary users should not be able to change role
A plain user was able to edit their profile to make themselves a power-user.

This should not be possible.

I'm pretty sure a plain user should not be able to edit the Users table at all.

TahaKhanAbdalli:
fixes #978


IanMayo (approved):
Accepted via code inspection.

PR #989, opened 2024-01-19
field for id-Dept, warning massage implementd

### LINKED ISSUE #954: Introduce `Edit` for `Departments`
Departments aren't expected to get a lot of use.

But, we should provide `Edit` form for the department.

The create and edit pages should allow edit of id, name and active (since sometimes `id` is actually an accepted shortening of the full name)

TahaKhanAbdalli:
fixes #954

### DIFF
--- a/src/pages/ReferenceData.tsx
+++ b/src/pages/ReferenceData.tsx
@@ -86,13 +86,15 @@ export default function ReferenceData(): React.ReactElement {
   const rarelyUsedRoutes = [
     { path: '/organisation', title: 'Organisation' },
     { path: '/protectiveMarking', title: 'Protective Marking' },
+    { path: '/mediaType', title: 'Media Type' },
+    { path: '/vault', title: 'Vault' }
+  ]
+
+  const rarelyThirdRowUsedRoutes = [
     { path: '/catCode', title: configData?.catCode ?? 'Cat Code' },
     { path: '/catHandle', title: configData?.catHandle ?? 'Cat Handle' },
     { path: '/catCave', title: configData?.catCave ?? 'Cat Cave' },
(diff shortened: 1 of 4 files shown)


IanMayo:
ok, I did this testing:
1. tried to change id for existing department -> error reported, edit not made
2. create new department  -> worked
3. edit new department to edit id  -> worked: error was reported, but new id was saved
4. assign batch to new department -> worked
5. now try to edit department id -> error reported, edit not made
6. assign batch away from new department -> worked
7. now try to edit department id -> worked: error was reported, but new id was saved

So, we've inherited this dodgy piece of database design.  This PR allows id values to be edited.  

If we can prevent the error from being reported, or change the error, or warn the user about the error [*] before doing `Save` on this forms if the`id` has changed - then I think it's ok.

[*] Maybe a popup saying: Changing the id of a row may cause an `Element does not exist` error to pop up. Please ignore this error, but be sure to verify that the id value has been updated.

TahaKhanAbdalli:
Okay.

IanMayo:
Strangely, when I edit the id for a department that is in use, I get your popup warning, but then react-admin silently fails. No warning is shown, and the new id isn't saved.

https://github.com/DeepBlueCLtd/RCO/assets/1108513/f012f509-b69c-42b8-93f8-6ce457d91770

TahaKhanAbdalli:
image
I made changes according to this.

IanMayo:
> I made changes according to this.

Sure - but... I get your popup warning that an error may get shown, but then the error isn't shown, so I think everything worked.  Ah, I think I'll just change the popup warning.

IanMayo (approved):
Tested and working

PR #1007, opened 2024-01-24
catCode/catCave/catHandle id is editable now

### LINKED ISSUE #961: Need the ID to be specified for new values of catCode, catCave, catHandle
The ID field for these tables contains an abbreviation of the `Name` - so it is used in presenting the data.

So, when we `Create` one of the values, we should allow the `id` to be specified.

Also, we should have an `Edit` page, which allows the `id`, `name` and `active` to be edited.

Hmm, editing an `id` after it has been created could cause problems if it has been used as an FK anywhere.  I'm pretty sure SQLite will catch that happening, and throw an error.  We should catch errors from saving an edited one of these and report "Error: potential break to referential integrity. Please obtain database support".

Note: we may need to show a warning like this one: https://github.com/DeepBlueCLtd/RCO/issues/954#issuecomment-1886817936

TahaKhanAbdalli:
fixes #961

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -126,10 +126,3 @@ export const cosmeticLabels = {
   [R_DEPARTMENT]: 'Department',
   [R_PROTECTIVE_MARKING]: 'Protective Marking'
 }
-
-export const resourcesWithListPage = [
-  R_DEPARTMENT,
-  R_CAT_CODE,
-  R_CAT_CAVE,
-  R_CAT_HANDLE
-]
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -126,3 +126,10 @@ export const cosmeticLabels = {
   [R_DEPARTMENT]: 'Department',
   [R_PROTECTIVE_MARKING]: 'Protective Marking'
 }
+
+export const resourcesWithListPage = [
+  R_DEPARTMENT,
+  R_CAT_CODE,
+  R_CAT_CAVE,
+  R_CAT_HANDLE
+]
--- a/src/resources/reference-data/ReferenceDataForm.tsx
+++ b/src/resources/reference-data/ReferenceDataForm.tsx
@@ -22,7 +22,7 @@ import FlexBox from '../../components/FlexBox'
 import { Modal, Typography } from '@mui/material'
 import { Box } from '@mui/system'
 import { useConfigData } from '../../utils/useConfigData'
-
+import { resourcesWithListPage } from '../../constants'
 const schema = yup.object({
   name: yup.string().required()
 })
@@ -42,6 +42,7 @@ const ChangeId = ({ handleClose, name }: Props): React.ReactElement => {
     boxShadow: 24,
     p: 4
   }
+
   const record = useRecordContext()
   const { getValues } = useFormContext()
   const notify = useNotify()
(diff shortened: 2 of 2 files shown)


src/resources/reference-data/ReferenceDataForm.tsx
```diff
@@ -143,9 +146,16 @@ export default function ReferenceDataForm(
     name?: string
   }): React.ReactElement => {
     const createRecord = useCustomid()
+    const validName = name ?? ''
+    const resourcesWithListPage = [
```
IanMayo, line 150: we should just declare this constant once, please.

src/resources/reference-data/ReferenceDataForm.tsx
```diff
@@ -201,13 +211,29 @@ const FormContent = ({
...
+    constants.R_CAT_HANDLE,
+    constants.R_DEPARTMENT
+  ]
 
-  const warningTextForId =
-    'Warning: Editing the id of a Department that is in use may lead to data corruption.  The id of a department must not be modified if data has been assigned to that department.'
+  const resourceName =
+    name === constants.R_DEPARTMENT
+      ? 'department'
+      : name === constants.R_CAT_CODE
```
IanMayo, line 225: The strings for `code`, `cave` and `handle` have to come from the `ConfigData`, please.

TahaKhanAbdalli: Oky let me check.

src/resources/reference-data/ReferenceDataForm.tsx
TahaKhanAbdalli: okay. Sorry I haven't seen this massage before. Let me fix it now. Please confirm you are talking about `const validName = name ?? ''` ?

IanMayo: No, I mean the array of resource names in `resourcesWithListPage`.  It contains business logic, but is declared several times.

TahaKhanAbdalli: Oky.

src/constants.ts
```diff
@@ -126,3 +126,10 @@ export const cosmeticLabels = {
   [R_DEPARTMENT]: 'Department',
   [R_PROTECTIVE_MARKING]: 'Protective Marking'
 }
+
```
IanMayo, line 129: Oh, I think we've lost a comment from yesterday - probably my mistake.  Fro the `resourcesWithListPage` const, since it is only used in `ReferenceDataForm`, can we declare it in that file?

(happy to discuss...)

IanMayo:
> Sorry @ian, I didn't understand what you are trying to say.

That's ok :-)  Do you remember why we're directing to `list` rather than `show` when we update these 4 resource types?

TahaKhanAbdalli:
> > Sorry @ian, I didn't understand what you are trying to say.
> 
> That's ok :-) Do you remember why we're directing to `list` rather than `show` when we update these 4 resource types?

Alright, Actually, on the show page, it was showing an error. That's why I redirected to the list instead of the show page.

IanMayo (approved):
Tested and working

src/resources/reference-data/ReferenceDataForm.tsx
IanMayo: Done.

src/constants.ts
IanMayo: Done.

PR #1010, opened 2024-01-24
cannot create new user issue resolved

### LINKED ISSUE #1004: Cannot create new user
When I try to create a new user, I get this error:
image

I've had a quick look at the code, but can't see where the `active` field is being generated.  Note: while we have `active` for other resource types, for users this concept is managed via the `departureDate`

TahaKhanAbdalli:
fixes #1004

### DIFF
--- a/src/resources/users/UserForm.tsx
+++ b/src/resources/users/UserForm.tsx
@@ -12,7 +12,7 @@ import EditToolBar from '../../components/EditToolBar'
 import { Typography } from '@mui/material'
 import { rolesOptions } from '../../utils/options'
 import FlexBox from '../../components/FlexBox'
-
+import { DateTime } from 'luxon'
 const schema = yup.object({
   name: yup.string().required(),
   role: yup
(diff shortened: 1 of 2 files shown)


IanMayo:
Thanks @TahaKhanAbdalli - do you understand where the issue came from?  I'm inquisitive about where the `active` field came from.

TahaKhanAbdalli:
The problem was with the exporting object from the `user/index.tsx` file. This object manages routes, but I forgot to include the create function in it. As a result, it was redirecting to the `reference data` create form by mistake.

IanMayo:
Thanks @TahaKhanAbdalli 

I should have noticed we were looking at the wrong form, duh!

Are you able to give the date a default value of 10 years in the future?

TahaKhanAbdalli:
If the date you mentioned is `departedDate` it's already 10 years in the future.

IanMayo:
> If the date you mentioned is `departedDate` it's already 10 years in the future.

Yes, `departedDate`. I just have an empty date field:
image

TahaKhanAbdalli:
Oky, Let me check.

TahaKhanAbdalli:
> > If the date you mentioned is `departedDate` it's already 10 years in the future.
> 
> Yes, `departedDate`. I just have an empty date field: image

Done

IanMayo (approved):
Tested and working

PR #1012, opened 2024-01-25
recipient name not shown in receipt note issue resolved

### LINKED ISSUE #1001: Recipient name not shown in Receipt Note
For the receipt note form, we don't show the name of the recipient. We should:
image

It's also an issue for the `Hastener note`:
image

TahaKhanAbdalli:
fixes #1001


IanMayo:
Thanks, one layout change.   Could you remove "Recipient name", and put the value at the start of the "To" field please?  So, the value in "To:" will be the name then a comma, then the address, please.

IanMayo (approved):
Tested and working

TahaKhanAbdalli:
Sorry @ian, I was not on my system.

PR #1013, opened 2024-01-25
invalid count on muster list issue resolved

### LINKED ISSUE #1011: Invalid count on muster list
The list of items shown on a Muster List is invalid.

In the below report, it says there are 8 items, but only 7 are listed. The valid value of 7 is shown in the count at the bottom left, and the breakdown by protective marking is also valid.

image

TahaKhanAbdalli:
fixes #1011


IanMayo (approved):
Tested and working

PR #1017, opened 2024-01-26
delete removed from address edit.

### LINKED ISSUE #959: Remove `Delete` from `AddressEdit`
We should remove the `Delete` button from the `AddressEdit` page.  It currently (silently) fails.

Aah, this could be useful.  I don't like the fact that it silently fails.

Can we try to catch the error being thrown, and show it in a popup?

Then we should remove the `Delete` button.

TahaKhanAbdalli:
fixes #959


IanMayo (approved):
Tested and working

PR #1018, opened 2024-01-29
in batch list and item list the porject is not click able now.

### LINKED ISSUE #995: Remove `click` for `Project` in `BatchList` and `ItemList`
The `Project` field is shown as clickable in `BatchList` and `ItemList`.  It would be better if it wasn't clickable, so clicking on that field marks the row as `selected` rather than navigating to `ProjectEdit`/`ProjectShow`

TahaKhanAbdalli:
fixes #995


IanMayo (approved):
Tested and working

PR #1019, opened 2024-01-29
Now the value of project coming from config data file.

### LINKED ISSUE #1008: In `BatchList` we should use the config data field for `Project`
We provide filters for the BatchList page.  It contains a `project` filter.  The name for this filter should actually come from the ConfigData (I think it's `projekt` in our sample data).

Note: while we are in there, the `Platform eq` filter should just be named `Platform`

Oh, and lastly, could `Platform` and `Active Platforms` be moved down to be near `Project`.  Oh, but the name of the `Project` filter should come from from `configData` (as above).

image

TahaKhanAbdalli:
fixes #1008

### DIFF
--- a/src/resources/batches/BatchList.tsx
+++ b/src/resources/batches/BatchList.tsx
@@ -110,22 +110,10 @@ export default function BatchList(): React.ReactElement {
       dataPickerProps={{ views: ['year'] }}
     />,
     <SourceInput
-      reference={constants.R_PLATFORMS}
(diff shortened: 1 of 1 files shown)


IanMayo:
Fixes the issue, but there's another mild issue, while you're working with this code:
image


Can we lose the ` eq` from the label shown in the platform filter please?

IanMayo (approved):
Tested and working

PR #1021, opened 2024-01-30
export button removed

### LINKED ISSUE #990: Remove Export Buttons
A number of `List` views include an `Export` button.

This button is problematic, since it is exporting raw values, it doesn't fill in the lookup fields for Foreign Keys.

But, a user can copy/paste directly from the displayed data into Excel or Word - which does contain the populated lookup values.

So, a quick fix is to remove the `Export` button form all list views.

TahaKhanAbdalli:
fixes #990


IanMayo (approved):
Tested and working.

PR #1022, opened 2024-01-30
need to sort batchlist by reference issue resolved

### LINKED ISSUE #996: Need to sort `BatchList` by `Reference Number`
We can't currently sort `BatchList` by `Reference` (it's not clickable).  We should be able to.

TahaKhanAbdalli:
fixes #996

### DIFF
--- a/src/resources/batches/BatchList.tsx
+++ b/src/resources/batches/BatchList.tsx
@@ -180,7 +180,7 @@ export default function BatchList(): React.ReactElement {
         <FunctionField<Batch>
           label='Reference'
           render={(record) => `${record.vault?.[0]}${record?.batchNumber}`}
-          sortBy='id'
+          sortBy='batchNumber'
         />
         <SourceField<Batch>
           source='department'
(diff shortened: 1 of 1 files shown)


src/resources/batches/BatchList.tsx
```diff
@@ -180,6 +180,7 @@ export default function BatchList(): React.ReactElement {
         <FunctionField<Batch>
           label='Reference'
           render={(record) => `${record.vault?.[0]}${record?.batchNumber}`}
+          sortBy='id'
```
IanMayo, line 183: I'm pretty sure the above should be sorted by `batchNumber`

IanMayo (changes requested):
The data doesn't look like it is being sorted by batch reference.

src/resources/batches/BatchList.tsx
TahaKhanAbdalli: If it is okay for. Can I fix it tomorrow?

IanMayo: Completely fine :-D.

TahaKhanAbdalli: Thank you. @IanMayo

IanMayo (approved):

Issue #1002, opened 2024-01-23
Display time for `start` and `end` for `ItemShow`


IanMayo:
We store time for `start` and `end` for Items, as in this `ItemList`:
![image](https://github.com/DeepBlueCLtd/RCO/assets/1108513/1cb3823e-549c-42f9-8d3e-32beb8db529c)

But, in `ItemShow` we only show the data element, the time element is missing.  We should include the time value.
![image](https://github.com/DeepBlueCLtd/RCO/assets/1108513/00089d50-6e16-4693-907a-6c35910cd1d4)

PR #1023, opened 2024-01-31
single quote issue resolved

### LINKED ISSUE #1003: Single quote in text field throws error
If we try an give a user the name `Rory O'Brien` we get a SQL error.

We need to be able to use this character in names and addresses.

Once we know what substitution we need to make, we should probably produce a `sanitise text` method.

Note: this _may_ be a `soul-cli` issue.

### LINKED ISSUE #1005: Address accepts a quote on `Create` but not on `Edit`
I can `create` a new address that contains a `'` character: `22 Dumb'rain street`

But, if I edit then save that address `soul` throws an error:
image

Note: I get the same issue for items.  I can include a `'` in the `consec` for an item being added to a batch, but If I subsequently edit that item, it will not save.

### LINKED ISSUE #1006: Including a `'` in free text search throws error
The search functionality throws an error if we include `'` in the free text search.

This is a character that can legitimately occur in names and addresses, so it should be allowable.

TahaKhanAbdalli:
fixes #1006 
Fixes #1003 
Fixes #1005

### DIFF
--- a/src/providers/dataProvider/index.ts
+++ b/src/providers/dataProvider/index.ts
@@ -58,6 +58,9 @@ const getConfigData = (): { configData: () => Promise<ConfigData | null> } => {
     }
   }
 }
+/** SQLite throws an error on a single quote. Swap for double-quotes.
+ * Note: we don't need to reverse this for received data, it is displayed correctly.
+ */
 const sanitizeCode = (data: any): string | any => {
   if (typeof data === 'string') {
     return data.replace(/'/g, '\'\'')
--- a/src/providers/dataProvider/index.ts
+++ b/src/providers/dataProvider/index.ts
@@ -58,6 +58,12 @@ const getConfigData = (): { configData: () => Promise<ConfigData | null> } => {
     }
   }
 }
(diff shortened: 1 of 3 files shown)


TahaKhanAbdalli:
Issues #1005 and #1002 are also addressed in this same PR.

IanMayo:
Thanks @TahaKhanAbdalli . But, do we need to swap the two quotes for one when data comes back from `Soul`?

(build not complete yet - so I can't test it myself)

TahaKhanAbdalli:
> swap
No, There is no need to swap two quotes for one.

IanMayo:
A couple of linter issues, please @TahaKhanAbdalli 
image

IanMayo:
> No, There is no need to swap two quotes for one.

I can see that we don't need to _Fix_ the two quotes, and that data renders ok.  Can you explain why we haven't had to do that?

IanMayo (approved):
Tested and working.

IanMayo:
Hello @TahaKhanAbdalli - I've tagged this with the 3 issues that I think it fixes. Would you mind double-checking please?

image

PR #1025, opened 2024-02-01
batchlist cannot filter for active platform issue resolved

### LINKED ISSUE #1009: BatchList - cannot filter for ActivePlatforms
For the `BatchList` page, we have a filter called `Active Platforms`.

This filter reduces the Batches to those from Active Platforms.

But, when it is applied, the REST call is mangled, since it queries the `Batches` table, not `Batch`.

image

TahaKhanAbdalli:
Fixes #1009

### DIFF
--- a/src/resources/batches/BatchList.tsx
+++ b/src/resources/batches/BatchList.tsx
@@ -68,7 +68,7 @@ const PlatformFilter = (props: PlatformFilterType): React.ReactElement => {
   const { data } = useGetMany(reference, { ids: platformIds })
   useEffect(() => {
     if (data != null) {
-      const filteredData = data.filter((d) => d.active === true)
+      const filteredData = data.filter((d) => !!d.active)
       setFilters(
         {
           ...displayedFilters,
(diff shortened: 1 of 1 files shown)


src/resources/batches/BatchList.tsx
```diff
@@ -61,7 +61,7 @@ interface PlatformFilterType {
 }
 
 const PlatformFilter = (props: PlatformFilterType): React.ReactElement => {
-  const { data: batches } = useGetList('batches')
+  const { data: batches } = useGetList('batch')
```
IanMayo, line 64: Hello @TahaKhanAbdalli - the cause of this issue is that we weren't referring to the `R_BATCHES` const.  Let's use it here, please.

IanMayo (changes requested):
Tested and working. Just one _taste_-related change.

src/resources/batches/BatchList.tsx
```diff
@@ -68,7 +68,7 @@ const PlatformFilter = (props: PlatformFilterType): React.ReactElement => {
   const { data } = useGetMany(reference, { ids: platformIds })
   useEffect(() => {
     if (data != null) {
```
IanMayo, line 70: Hello @TahaKhanAbdalli - the `ActivePlatform` filter wasn't working for me.  I changed the test to use a _logical_ true value, rather than an explicit `true` - and it worked.

Are you ok with it?

IanMayo (approved):
Tested and working.

PR #1027, opened 2024-02-02
button to reset displayed columns issue resolved

### LINKED ISSUE #991: Button to reset displayed columns
Users can choose which columns to display in most lists, they do this to tailor the content to the job in hand.

It is correct that RCO remembers which columns the user wants to view.

But, we should introduce a button (somewhere) to reset the currently displayed columns to the original default ones.

TahaKhanAbdalli:
Fixes #991


IanMayo (approved):
Tested and working

PR #1028, opened 2024-02-05
offer cosmetic clues of working on test server issue resolved

### LINKED ISSUE #1026: Offer cosmetic clues of working on test server
The software-testing guy reported a `Fail` because it wasn't easy to tell if he was on a test server or a publication server.

Since VAL is unaware of the infrastructure it is installed upon I couldn't really think of a way of resolving this, and once testing was complete the issue would go away anyway.

But, today the clients and I discussed wider use of the test server, by a slightly wider body of users.

So, I think this has become more important.

Could you have a think about this please?  How about if:
1. We have VAL code that inspects the window URL, and looks for the text `VAL-DEV`.  
2. If that text is present, then the blue header bar goes to dark red, and the [VAL] in the title bar is changed to [VAL-DEV]

How does that sound?

Note: for this, we wouldn't just look for `VAL-DEV` in the server part of the URL, we'd look in the whole of the window-URL.  So, to test this, we'd just to a free-text search for `VAL-DEV`, and that would trigger the UI colors to change.  I think.'..

**Update:** user requested that we also have suitable clarifying signage on the login/welcome screen to.  Have demoted from `bug` to `enhancement` as this fix isn't urgent.

TahaKhanAbdalli:
Fixes #1026

### DIFF
--- a/.env
+++ b/.env
@@ -2,5 +2,4 @@
 VITE_DATA_VERSION="2"
 # change this value to wipe cached UI prefs (including filters and visible columns)
 VITE_APP_VERSION="1"
-# set this to true to cosmetically style this instance as a test/Development server
-VAL_DEV = false
+
--- a/.env
+++ b/.env
@@ -2,5 +2,5 @@
 VITE_DATA_VERSION="2"
 # change this value to wipe cached UI prefs (including filters and visible columns)
 VITE_APP_VERSION="1"
-
-
+# set this to true to cosmetically style this instance as a test/Development server
+VAL_DEV = false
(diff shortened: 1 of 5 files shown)


IanMayo:
I'm running this locally.   It runs in `dev` styling when prop is passed via existing `yarn dev` command.

But, when I run via `yarn serve`  it doesn't pick up the value from `.env`.

Aah, hold on.  The `.env` gets picked up in `yarn build`.  So, I _guess_ the IT guys will have to do two separate build processes on the hosting/staging machine.

Can you think of a way of handling this which does not require `yarn build` after editing `.env`.  I _guess_ it could be on the command line, right?

IanMayo (approved):
Tested and working

PR #1034, opened 2024-02-19
ctrl-click handler applied on viewItem icon

### LINKED ISSUE #999: After editing an item, filters are lost
A user filtered the items to a specific set.

The user then wants to make a number of changes to these items.

But, after editing the first, when they return to `Live Items` the filters are lost.

I see two possible solutions for this:
- persistent filters
- use some `special` LiveItems url when going to `LiveItems`
- change the `ViewItem` icon into a clickable URL.  Then users can right-click and `open in new tab`. That means the original listing is unchanged.  Aah, or add `ctrl-click` handler to this icon which opens the link in a new tab (if possible)

TahaKhanAbdalli:
Fixes #999

### DIFF
--- a/src/components/DatagridConfigurableWithShow.tsx
+++ b/src/components/DatagridConfigurableWithShow.tsx
@@ -21,9 +21,13 @@ const PreviewButton = ({
   const redirect = useRedirect()
   const createPath = useCreatePath()
 
-  const handleClick = (): void => {
-    const path = createPath({ resource, type: 'show', id })
-    redirect(path)
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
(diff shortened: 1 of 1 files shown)


IanMayo:
Thanks @TahaKhanAbdalli  - but it isn't working for me (macos).  None of ctrl, command or alt open it in a new tab.  Hmm, it's also _hidden-behaviour_.  If we make it into a proper HTML link then the cursor changes, plus if users right-click on it, they get "open in new tab" (as well as bomb-proof ctrl-click behaviour).

Could you investigate showing it as a proper hyperlink please?

TahaKhanAbdalli:
Okay Let me check.

IanMayo (approved):

PR #1038, opened 2024-02-21
List pages missing title issue resolved

### LINKED ISSUE #1036: `List` pages missing titles
Hmm, this issue slipped through the cracks, too.

For `List` pages we don't show a title:
image

That's not actually a big issue in the above page, since the resource name is selected on the left-hand side.

But, for `Reference`  pages it's not obvious:
image

So, we should introduce titles for these pages, please.    The title can just be the collective name for the resource (I _think_ it's ok to just put an `s` on the end of the term you defined in the `resource name array` in PR #1035

TahaKhanAbdalli:
Fixes #1036


IanMayo (approved):
Good fix, thanks @TahaKhanAbdalli

PR #1039, opened 2024-02-21
Id field added in create form

### LINKED ISSUE #1032: Need to provide `id` at `create` for cat, code, handle
Due to patterns in the legacy data, the `id` is used as the short-version of the label for `cat-Code`, `cat-Cave` and `cat-Handle`.

While the id value is automatically generated, we allow the users to edit it after creation.

We should actually let them set the value in the `Create` form.

TahaKhanAbdalli:
fixes #1032


IanMayo (approved):
Tested and working

PR #1045, opened 2024-02-26
delete button removed from address list

### LINKED ISSUE #1044: Need to remove `Delete` button from `AddressList`
If we select an address from the list, we are offered a `Delete` button.  It cosmetically removes the address, but the address is restored on refresh.

We should remove this disable button

image

TahaKhanAbdalli:
fixes #1044

### DIFF
--- a/src/resources/addresses/AddressList.tsx
+++ b/src/resources/addresses/AddressList.tsx
@@ -17,7 +17,7 @@ export default function AddressList(): React.ReactElement {
       <Typography variant='h5' fontWeight={'bold'} style={{ padding: ' 15px' }}>
         Addresses
       </Typography>
-      <Datagrid rowClick='show'>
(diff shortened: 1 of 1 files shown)


src/resources/addresses/AddressList.tsx
```diff
@@ -17,7 +17,7 @@ export default function AddressList(): React.ReactElement {
       <Typography variant='h5' fontWeight={'bold'} style={{ padding: ' 15px' }}>
         Addresses
       </Typography>
-      <Datagrid rowClick='show' bulkActionButtons={<></>}>
+      <Datagrid rowClick='show' bulkActionButtons={false}>
         <TextField<Address> source='fullAddress' />
```
IanMayo, line 21: Hello @TahaKhanAbdalli - I did some googling, and found that the above removes the ability to select items, since that's no longer relevant.

IanMayo (approved):

PR #1050, opened 2024-03-07
remarks in destruction is now optional

### LINKED ISSUE #1047: Remarks in Destruction should be optional
The `Remarks` field is currently mandatory for `Destructions` / `Edit`.  It should be optional.

**Update**: we need to make remarks optional for `CREATE`, too.  The field is nullable in the schema, so that's not preventing `Create`.  I _think_ it's because the form needs to be made `dirty` before `Create` works.

TahaKhanAbdalli:
fixes #1047


IanMayo (approved):
Tested and working.

PR #1054, opened 2024-03-14
1052 update schema for new auth - DO NOT MERGE

### LINKED ISSUE #1052: Auth: 0.1 - Update users
We have to make some database updates to prepare for adopting soul-auth.

I _Think_ this is the logical flow:
1. Refactor `users` to `_users`
    a. manually remove relevant foreign key constraints in SQLite
    b. rename table
    c. manually re-create foreign key constraints
    d. rename table in T/S
    e. rename table name used in mock dataset
2. Rename columns in _users: 
    a. rename `password` to `hashed_password`
    b. rename `staffNumber` to `username`
    c. introduce `is_superuser` boolean field, set to false by default
    d. introduce `salt` string field, set to blank by default
    e. update field names in T/S (and in the T/S definition for user)
    f.  update generation code for mock dataset
3. Check app & auth still work as expected

IanMayo:
Fixes #1052 

Note: this is imcomplete.   The app now tries to access the `_users` table, but the resource is now called `_users` too.  Ideally, the resource would still be `user`.

**Important**: we won't use this branch. We'll use @TahaKhanAbdalli 's version #1055


IanMayo:
Drop this, it has been fixed separately

PR #1055, opened 2024-03-14
#1052 update user table

TahaKhanAbdalli:
This is a good piece of work. Well done Taha.

Some feedback:
- in the left-hand list of pages it says `users`, it should be `Users`
- I think you need to do some refactoring of `staffNumber` to `username` in the TS code


src/providers/authProvider/permissions.ts
```diff
@@ -28,9 +29,38 @@ const permissions: Record<UserRole, ResourcePermissions> = {
 
 export default permissions
```
IanMayo, line 31: Hello @TahaKhanAbdalli - I think VAL can be simpler than this. I don't think VAL needs to think about roles or permissions. VAL can ignore these. I think we can remove the concept of permissions from VAL. We just make the request of Soul, and either Soul will return result, or return an error indicating "unauthorised access".

Aah, hold on.  The client-side does need to know about the current user permissions, since we need to know whether to provide create/edit buttons.  Hmm, we _could_ hard-code those in T/S (since that is how we did it previously).  Then, Soul-auth is a `fall-back` - only allowing REST interactions that the user is authorised to do.

Hmm, I think our clients would probably prefer this to be configured in SQLite, than hard-coded in T/S.

So, that's an argument for accessing roles and permissions in the way you've been doing.

Well done, Taha - carry on. 🤣 .

Any thoughts?

TahaKhanAbdalli: Actually, we are rendering our UI according to `permissions`. So that's why we are doing this.

src/providers/authProvider/index.ts
```diff
@@ -18,7 +18,7 @@ import axios, { isAxiosError } from 'axios'
 import bcrypt from 'bcryptjs'
 
 export const getUser = (): _Users | undefined => {
-  const encryptedUser = getCookie(constants.TOKEN_KEY)
+  const encryptedUser = getCookie('accessToken')
```
IanMayo, line 21: @TahaKhanAbdalli - we use this string here and at line 38.  Could we use a constant instead, please?

src/providers/authProvider/index.ts
```diff
@@ -37,18 +35,24 @@ const setToken = (token: string): void => {
...
-  // removes it
-  console.log(name)
-  document.cookie = `${name}=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/;`
+const removeCookie = (): void => {
+  document.cookie =
+    'accessToken=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/;'
+
+  document.cookie =
+    'refreshToken=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/;'
+}
```
IanMayo, line 52: is there a reason why we set the cookie twice?  If there is, could you add a comment please?

src/providers/authProvider/index.ts
```diff
@@ -84,11 +88,11 @@ const updateLockouAttempts = async (
   })
 }
 
-const fetchUser = async (username: string) => {
+const fetchUser = async (username: string): Promise<any> => {
   const user = await axios.get(
     `http://localhost:8000/api/tables/_users/rows?_filters=username:${username}`
```
IanMayo, line 93: we can't use a hard-coded URL - it may be served on another port.

src/providers/authProvider/permissions.ts
```diff
@@ -29,28 +29,30 @@ const permissions: Record<UserRole, ResourcePermissions> = {
...
 }
 
-const mapPermissions = (permissions: any) => {
+const mapPermissions = (permissions: any): ResourcePermissions => {
   const mappedPermissions = permissions.reduce((acc: any, permission: any) => {
     acc[permission.table_name] = {
-      read: permission.read,
-      write: permission.create,
-      delete: permission.delete
+      read: permission.read === 'true',
```
IanMayo, line 49: are these values really stored as strings?  Normally booleans are stored as `1` or `0` on SQLite.

src/providers/authProvider/index.ts
IanMayo: I'm pretty sure this hard-coded path is causing the `yarn test` failure: https://github.com/DeepBlueCLtd/RCO/actions/runs/8374624709/job/22930278394

TahaKhanAbdalli: When we wrote the test case it was fetching permissions from the frontend but now as the permissions logic is shifted to backend, therefore it doesn't find the server running to make api call to fetch permissions therefore the connection error.

IanMayo: ok, I guess we can remove the test, since  the capability is no longer provided by RCO/VAL.

IanMayo (approved):
Good fixes, thanks.

PR #1066, opened 2024-03-25
filter on batch list displayed in vertical form issue fixed

### LINKED ISSUE #1059: 5. Filters on the batch view page are displayed as a vertical column,unlike on other pages
I suggest we implement this quick fix, since it is reducing the number
of batches displayed on screen.

TahaKhanAbdalli:
fixes #1059


TahaKhanAbdalli:
I have noticed the same issue in other resources too `Live Items`, `All Items` and `Batch List`. But it is fixed now.

IanMayo (approved):
Tested and working

PR #1067, opened 2024-03-26
Logging in as a user with no role gives blank page issue resolved

### LINKED ISSUE #1062: 1.  Logging in as a user with no user type set gives a blank page
This functionality will move from `VAL" to "Soul" in the "Soul-Auth"
update.  Once the transition is nearing completion we should test this
issue isn't present.

@TahaKhanAbdalli  - as we're adopting Soul-Auth, could you please check that VAL doesn't fall over if it encounters a user without a role?

TahaKhanAbdalli:
fixes #1062


IanMayo (approved):
Solution is logical.

PR #1068, opened 2024-04-03
1056 password reset

### LINKED ISSUE #1056: Auth: 3.x Password reset
When users forget their passwords, we need power-users to be able to reset their password.

Currently in VAL, when the power-user is looking at a User page, they are presented with a "RESET PASSWORD" button. When this is clicked the password field for that user is cleared.

Then, when the user logs in (using username in name AND password fields), they are directed to a "Change Password" page - and forced to update their password.

But, now we are using Soul auth, we are hindered because Soul requires the current password to be provided when setting a new password.   The RCO client doesn't have the existing password at the time of "Change Password", so we can't support that flow.

Instead, we are going to provide a "force change password" end-point, just for RCO.  When this is called, the backend will verify this is a power-user, and verify it's not the password for the current user that is being modified.  If valid, then the new password is assigned for that user.  A "login-cutoff-time" is then assigned, 60 minutes into the future.   So, a power-user can set the password for a user, then write it on a piece of paper.  That user then has 60 minutes to login using that password, at which point they will be forced to change their password (which also clears the 60 minute timer), 

Hmm, for this `60 minute` timer I guess we could add a `updateBefore` field.  When a power-user has force-changed a password, this is set 60 minutes into the future.  When there is a value in this field for a user, we display a prominent "Password update pending" field in the UI (somewhere).  Whenever a user changes their password, the `updateBefore` field is cleared.

At login, we check both `departedDate` and `updateBefore` fields when we process login details.

TahaKhanAbdalli:
fixes #1056


_devExtensions/editPassword-controller.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+        return value
+      })
+      .join(', ')
+
+    let values = `(${fieldsString}) VALUES (${valuesString})`
+    if (valuesString === '') {
+      values = 'DEFAULT VALUES'
+    }
+
```
IanMayo, line 54: I'm pretty sure we can modify the above code to use a const for values, something like:
```
   const values = valuesString === '' ?  'DEFAULT VALUES' : [longer-string-definition] 
```

_devExtensions/updateBefore-controller.js
```diff
@@ -0,0 +1,46 @@
...
+
+const getUserById = (db, userId) => {
+  const query = `
+    SELECT createdAt, createdBy, departedDate, id, is_superuser, lastUpdatedAt, lockoutAttempts, name, role, updateBefore, username
+    FROM _users
+    WHERE id = ?;
+  `
+  return db.prepare(query).get(userId)
+}
+
```
IanMayo, line 13: @TahaKhanAbdalli - the following string clears the `updateBefore` field. Should it be called `clearUserUpdateBefore`?

_devExtensions/editPassword-controller.js
IanMayo: Hold on.... We don't use the value of the field `values`

_devExtensions/updateBefore-controller.js
TahaKhanAbdalli: Sure

package.json
```diff
@@ -15,7 +15,7 @@
     "build": "tsc && vite build",
     "preview": "vite preview",
     "check-dist": "if [ ! -d dist ]; then echo '\\e[31mError: The \"dist\" folder does not exist. Please run \"yarn build\" first.\\e[0m'; exit 1; else echo 'Dist folder exists'; fi",
-    "serve:dev": "soul -d ./db/RCO2.sqlite -e $(realpath ./_devExtensions)",
+    "serve:dev": "node ./backend.js",
     "serve:large": "if [ ! -d dist ]; then echo '\\e[31mError: The \"dist\" folder does not exist. Please run \"yarn build\" first.\\e[0m'; else soul -d ./db/RCO2-large.sqlite -e $(realpath ./_extensions); fi",
```
IanMayo, line 19: hello @TahaKhanAbdalli - I don't remember why we introduced `backend.js` - could you explain it to me please?  I see that we just use it for `serve:dev`.

package.json
```diff
@@ -53,7 +53,7 @@
     "react-dom": "^18.2.0",
     "react-idle-timer": "^5.7.2",
     "react-router-dom": "^6.9.0",
-    "soul-cli": "^0.6.1",
+    "soul-cli": "0.7.3",
```
IanMayo, line 56: is it worth including the up-arrow, so we get soul updates automatically?

src/ChangePassword.tsx
```diff
@@ -63,25 +65,25 @@ const ChangePassword = ({
...
-  }
+  //   if (
+  //     lastUpdatedAt !== null &&
+  //     lastUpdatedAt !== '' &&
+  //     !isDateNotInPastDays(lastUpdatedAt, 1)
+  //   )
+  //     throw new Error(
+  //       'Password update not allowed. Please wait at least one day before updating your password again.'
+  //     )
+  // }
```
IanMayo, line 86: Do we still have the minimum password age testing?  Does it now happen somewhere else?

IanMayo:
Note: I've just watched the network calls.  The slug of data contains a `hashed_password` attribute, but it is carrying the raw password (not hashed).
Should it be hashed?

package.json
TahaKhanAbdalli: Okay, I will add this.

src/ChangePassword.tsx
TahaKhanAbdalli: We do not need this password age testing functionality. That's why I commented this. But I will remove this as well.

TahaKhanAbdalli:
> Note: I've just watched the network calls. The slug of data contains a `hashed_password` attribute, but it is carrying the raw password (not hashed). Should it be hashed? image

IanMayo:
Did you mean to close this PR @TahaKhanAbdalli ?

TahaKhanAbdalli:
> Did you mean to close this PR @TahaKhanAbdalli ?

Closed it by mistake

IanMayo:
Ok. this isn't password reset related, but it looks like Auth.   I tried to add items to a dispatch. As I did so, I got a login dialog.  Once I logged in, the items had not been added:
screencast

IanMayo:
Done some more testing:
1. When I selected `RESET Password` for Petunia, it immediately logged me out.  I'm the power-user looking after user data, it's not me that has to re-login again.  I _guess_ we are having some unexpected impact on the backend.

TahaKhanAbdalli:
Hello @IanMayo, there is an issue related to resetting the password. When we try to reset the password, it does not empty the `hashed_password` field in the database. This occurs because we are removing the `hashed_password` field from the payload in SOUL.
image

I have some solutions:
1. Update the SOUL.
2. Create a new custom API to remove the password.
3. Lastly, I believe we should remove the `Reset Password` option because we already have the `Edit Password` functionality, which is essentially the same.

IanMayo:
Yes - I agree. We can remove the `reset-password` now that we have the `set password` feature for power-users.

TahaKhanAbdalli:
We are also resetting the `lock attempts` to zero with this `reset password` button. So now, I will add that feature to `set password`. Is that okay?

IanMayo (approved):

PR #1070, opened 2024-04-08
Provide context-sensitive title to reset password dialog.

### LINKED ISSUE #1069: Dialog message for renew password
If a user hasn't logged in for a while, their password can expire.

When this happens the reset password dialog is shown.

image

But, the message refers to setting an initial password, which is confusing.  The title should refer to updating a password.

Note: one way of handling these multiple titles would be to switch `resetPasswordOpen` to being a string rather than a boolean. The field would be changed to `resetPasswordTitle`.  The modal would be shown when that was non-null, and `setPasswordTitle` would take a value of `Please provide an initial password` or `You password has expired, please provide an updated one`.

IanMayo:
Fixes #1069

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -336,7 +336,10 @@ function App(): React.ReactElement {
 
           if (!password) {
             setResetPasswordTitle('Please provide an initial password')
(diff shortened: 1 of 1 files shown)


IanMayo:
close and re-open, to wake up linter.

PR #1075, opened 2024-04-17
Adopt new field title, drop staffNumber

### LINKED ISSUE #1074: Loan Item not working, due to `staffNumber`
We have removed `staffNumber` from the database.  It has been changed to `username`.

We should reflect this in the front-end code.

IanMayo:
Fixes #1074


TahaKhanAbdalli:
Tested and Working.

PR #1078, opened 2024-04-18
1076 cannot add new batch items

### LINKED ISSUE #1076: Can't add new `Batch` items
I get this error when trying to add a new batch:
image

RCO is trying to POST to this resource:
image

Clearly the data is missing batch.id, since it hasn't been assigned yet:
image

SQLite throws this error:
image

I notice that the field isn't set to `AutoIncrement`
image

This can't be an issue for the clients, since they have created loads of batches.  It's prob just in our data.

TahaKhanAbdalli:
fixes #1076


IanMayo (approved):
Working perfectly

PR #1079, opened 2024-04-19
1058 filter of destruction on item list not showing complete results

### LINKED ISSUE #1058: 6. Filtering the item list by ‘destroyed’ doesn’t show all destroyed items
This issue should be resolved, since VAL isn't showing accurate results.

Steps to reproduce:
- login to mock instance
- view `All items`
- add filter for `Destroyed`
- only one item is shown

But, from SQLite it's clear that items 2 and 2019 have been destroyed.

I checked the `network` tab. It's showing strange behaviour.  There is a call to `richItems` that gets the two rows back. Then there is another call, which just gets the single one we see.

TahaKhanAbdalli:
fixes #1058

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -49,6 +49,7 @@ import SourceField from '../../components/SourceField'
 import { useLocation } from 'react-router-dom'
 import useItemList from '../../hooks/useItemList'
 import { useConfigData } from '../../utils/useConfigData'
+import NullUndefinedFilter from '../../components/NullUndefinedFilter'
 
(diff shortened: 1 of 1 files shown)


IanMayo (commented):
Thanks Taha - the solution works for my dataset

Can you give me an overview of what the issue was please?

TahaKhanAbdalli:
Hello @IanMayo, I first applied a `booleanFilter` to filter out destroyed items from the item list. The `booleanFilter` runs two queries: it first checks if the `destructionDate` is not null, and then it match the data which is not null with the database which causing incorrect result.

Now, I have applied the `NullUndefinedFilter`, which filters based on `destructionDate` not being null.

PR #1082, opened 2024-04-22
now user role in _users_roles table are editable.

### LINKED ISSUE #1080: Need to use soul-auth roles, not RCO
In our user table we have roles.

But, in the soul-auth permissions model, these are managed via entries in the `_user_role` table.

So, we need to change the UI of the `User` page to reflect this new model.  The UI will show the `roles` entries for this user, and  `power-users` should be able to change which `roles` a user has.

Hmm: `soul-auth` allows for multiple roles per user, so a set of permissions is `built-up`.  But, the permissions model we have implemented just uses one role instance for users and another for power-users, we don't accumulate them.  So, we _could_ "hide" the one-to-many relationship - i.e. Provide a drop-down of the two roles, and when it is modified, clear all `user_role` entries for that user, and just store the new one.

TahaKhanAbdalli:
fixes #1080


IanMayo:
I've tried playing with the `role` for user-2 (petunia).

I have this feedback:
- form doesn't go `dirty` if I change the role. I also have to change the name in order for `Save` to be enabled.
- we aren't able to `clear` the role for the user, the backend (silently) throws an error.
- this is another `show` form where the `Edit` can disappear off to the right. Can we left-align the UserShow actions please?

IanMayo:
Just one item of feedback. please @TahaKhanAbdalli 

When I change a user role, the audit log shows a password update and a user resource edit.  Neither other these actually happened, and the user role-change isn't recorded.

So, could we please:
- if the user role has changed, create an audit entry for the user (showing the previous role)
- if no other fields have changed, then let's not create an audit entry or PUT an updated record to the server.

src/resources/users/UserForm.tsx
```diff
@@ -26,34 +41,184 @@ const schema = yup.object({
...
+      return
+    }
+    const roleResponse = await axios.get(
+      `${BASE_URL}/api/tables/_users_roles/rows?_filters=user_id:${userId}`
+    )
+    const roleId = roleResponse.data.data[0]?.id
+    if (!roleId) {
+      return
+    }
+
```
IanMayo, line 83: I think we need more robust handling for the above block, to handle cases where legacy users don't have roles.  In this case, the roleId will be empty.  When this happens, we `POST` a new `_user_role` object with the `user_id` and the new `role id`.

This is how the admin staff will bring on board other users as adoption of VAL grows.

IanMayo: Fixed in #1085

IanMayo (approved):
Single issue identified (infinite loop if wrong password provided in change password), moving that to new "testing" issue.

PR #1084, opened 2024-04-25
refresh rendering issue resolved

### LINKED ISSUE #1083: Refresh rendering issue
I've been testing the access/refresh processes.  I set the atet to 10 seconds and rtet to 30 seconds (or something like that).  Then I repeatedly went forward and backwards between the "MediaTypeList" page and "MediaTypeShow" (I think).

If an access token has expired, the server returns an error message.

This gets rendered to the screen, and I believe ReactAdmin stops rendering the current screen
refresh_render

Is it possible to handle the refresh, then re-attempt to retrieve the content, so the screen is displayed properly?

Ideally we won't show the red error box for something in the routine access/refresh cycle, or have the `Something went wrong` UI element.

TahaKhanAbdalli:
fixes #1083

### DIFF
--- a/src/utils/helper.ts
+++ b/src/utils/helper.ts
@@ -216,7 +216,13 @@ axios.interceptors.response.use(
   (response: AxiosResponse) => response,
   async (error: AxiosError) => {
     if (error.response && error.response.status === 401) {
-      await handleRefreshToken()
+      try {
+        await handleRefreshToken()
+        const originalRequest = error.config
+        if (originalRequest) return await axios(originalRequest)
+      } catch (refreshError) {
+        return await Promise.reject(refreshError)
+      }
     }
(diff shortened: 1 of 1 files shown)


IanMayo:
The fix works perfectly, thanks @TahaKhanAbdalli 

It was simpler than I feared.

But, can you give me a quick explanation of how the fix works please?
image

TahaKhanAbdalli:
Sure @IanMayo. 
Previously, the interceptor was solely responsible for refreshing the token upon receiving a 401 error. Now, with the modification, not only does it refresh the token, but it also captures the `originalrequest` and retries it with the new token once the token refresh is successful.

IanMayo:
Great, thanks

IanMayo (approved):
Fix works perfectly.

PR #1085, opened 2024-04-25
Hotfix update role for legacy users

IanMayo:
Our old users do not have roles in the `_user_roles` table.  I identified a problem where we couldn't give them a role, since the code relied on finding the `_user_role`.

In this PR, if the existing `_user_role` isn't found, then it creates one.


TahaKhanAbdalli:
Tested and working

PR #1087, opened 2024-04-26
Temporary trial database upgrade

IanMayo:
Practice performing migration to new database structure.

Currently failing, since it thinks `20300` has no user permissions:
image


IanMayo:
Close this PR/branch. Will re-do trial of migration using updated VAL sources

PR #1091, opened 2024-04-30
fixed current password issue

### LINKED ISSUE #1090: Testing RCO with updated auth
The changes in support of the transition to Soul-Auth have been quite extensive, though negligible changes have been made to the asset management processes.

The testing should at least include:
* [x] atet/rtet expiry
* [x] password expiry
* [x] setting password
* [x] account expiry (user departed date)
* [x] cannot re-use existing password (I think I've seen this working correctly this week)
* [x] minimum password time (user cannot change password one minute after changing it)
* [x] ISSUE: if I provide the wrong current password when changing password I get stuck in an infinite loop.

TahaKhanAbdalli:
Fixes #1090 

Current password issue resolved.


IanMayo (approved):
Tested and working

PR #1093, opened 2024-05-01
1092 - single logic for update and change password first commit

### LINKED ISSUE #1092: Single logic for update/change password
Passwords get updated for two reasons:  
1. system triggers password update when password has expired, or when Power-User provides temporary password
2. user elects to change password

We provide custom logic for `1.`, and use Soul-Auth logic for `2.`.

But, the second strategy doesn't meet our password checking criteria.  We will modify `2` to use the existing logic for `1`, but adding a step to validate the current password.

* [ ] extend `api.js`  to introduce `validateCurrentPassword` method. This will receive the mainDb, the `userId` and the  `password` . It will hash the password and compare it with the current `hashed_password`  for the specified `userId`
, throwing an error if it is wrong
* [ ] extend insertPasswordRecord to check if a currentPassword is received, and then call this method before we do passwordValidationScehema.validate()
* [ ] modify changeAndUpdatePassword to call the insert-password end-point instead of change-password

TahaKhanAbdalli:
Fixes #1092

### DIFF
--- a/_devExtensions/api.js
+++ b/_devExtensions/api.js
@@ -15,6 +15,21 @@ const getIp = {
   }
 }
 
+const validateCurrentPassword = (db, userId, currentPassword) => {
+  const query = `SELECT hashed_password FROM _users WHERE id = ?`
+  const user = db.prepare(query).get(userId)
+  if (!user) {
+    throw new Error('User not found')
+  }
+  const isPasswordValid = bcrypt.compareSync(
+    currentPassword,
+    user.hashed_password
+  )
+  if (!isPasswordValid) {
+    throw new Error('Invalid current password')
+  }
+}
+
 const removeOldPasswords = (db, userId) => {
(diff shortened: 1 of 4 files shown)


_devExtensions/api.js
```diff
@@ -15,6 +15,21 @@ const getIp = {
   }
 }
```
IanMayo, line 17: Hello @TahaKhanAbdalli - when I was implementing this new feature I thought I was re-using the existing `validateCurrentPassword` method.  

I now realise that Github CoPilot write it for me.  I'm pretty amazed at that. It all looks valid to me, and works correctly - but I'd appreciate you giving it extra scrutiny please.

TahaKhanAbdalli:
Tested and Working

IanMayo (approved):
Taha has reviewed my implementation.

PR #1095, opened 2024-05-01
Correctly build URL in production

### LINKED ISSUE #1094: Workaround failed to fix issue
On the test server, login worked, but the REST call to get the user details failed, since it was trying to retrieve from `localhost`

The underlying reason for this was that the `.env` was missing the `process.env.API_BASE_URL_KEY`.  We test for this field in 4 places.

It would be my preference to not rely on the value of this field, since it would have to be maintained across test and prod servers, and included in every update.

I prefer the use of the `process.env.NODE_ENV === 'development'` test., which we use in 16 places.

I attempted to fix the test, but I got it wrong, I should have used the more verbose `if` construct.

IanMayo:
Fixes #1094


TahaKhanAbdalli:
Tested and Working

PR #1096, opened 2024-05-01
Hotfix update database schema

IanMayo:
The database schema has changed a lot with the new auth.  Update the schemas.


IanMayo:
Note: codacy reported some errors, but that is because it is verying TSQL. Our code is valid SQLite.

PR #1106, opened 2024-05-03
Modify serve script to extract token secret from the envrionment

### LINKED ISSUE #1104: Move secret token out of `yarn serve` command.
Having a `.env` file was complicating the run-time process by requiring an extra layer.

But, specifying the `-ts` parameter in the command to run the app means it is then visible in the `ps` process listing for the server. Their sys-admin sees this as a security problem.

We should move the `-ts` attribute back to the `.env` file, and restore the step to pull environment variables into the `yarn serve` command.

IIRC, we dropped the `backend.js` way of starting Soul, which meant we couldn't use the `.env` file any more.  I still don't like `backend.js`

Hmm, if we don't provide `-ts` in the command line, won't `Soul` pull it from the `.env` file anyway?

AbegaM:
### Modifications 
1. Modified the `serve` script so we can be able to read the `TOKEN_SECRET` config key from the `.env` file

Fixes #1104


IanMayo (changes requested):

IanMayo (approved):
Tested and working

PR #1115, opened 2024-05-22
removed reference-data for default user.

### LINKED ISSUE #1114: Don't offer a `default-user` the `Reference Data` page
I'm logged in as a `default user`.

I'm offered the `Reference Data` page in the left-hand menu.

We do not need to offer default users this level of information.

TahaKhanAbdalli:
fixes #1114


IanMayo (approved):
Tested and working

PR #1116, opened 2024-05-22
1113 no receipt note default user

### LINKED ISSUE #1113: Should not invite default user to accept receipt note
I'm logged in as a default user.

From the Welcome page I can see a list of Pending Receipt Notes.  If I click on the title, I am shown the list.

If I select one, I am invited to mark `Receipt Note Received`.  If I do that, the backend rejects the operation, and I get logged out.

This is evidence that the backend security is working, but the front-end shouldn't offer the action to a `default-user`.

TahaKhanAbdalli:
fixes #1113

### DIFF
--- a/src/providers/authProvider/permissions.ts
+++ b/src/providers/authProvider/permissions.ts
@@ -65,9 +65,15 @@ const mapPermissions = (
       write: true,
       delete: false
     }
-  } else {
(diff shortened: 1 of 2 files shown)


TahaKhanAbdalli:
Note: do not use this PR.
Use this one: 1113_no_receipt_note_for_default_user

PR #1117, opened 2024-05-22
no reciept notes for default_user

### LINKED ISSUE #1113: Should not invite default user to accept receipt note

TahaKhanAbdalli:
fixes #1113

### DIFF
--- a/src/resources/dispatch/DispatchList.tsx
+++ b/src/resources/dispatch/DispatchList.tsx
@@ -90,10 +90,10 @@ const BulkActions = (): React.ReactElement => {
       type: 'success'
     })
   }
-  const rolesThatCanEditPassword = ['rco-user', 'rco-power-user']
+  const rolesThatCanCreateReceiptNote = ['rco-user', 'rco-power-user']
   return (
     <FlexBox>
-      {user && rolesThatCanEditPassword.includes(user.userRole) ? (
+      {user && rolesThatCanCreateReceiptNote.includes(user.userRole) ? (
         <Button
           disabled={!showReceiptButton}
           onClick={receiptReceived as any}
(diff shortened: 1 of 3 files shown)


TahaKhanAbdalli:
Okay  let me check.

IanMayo:
Aah, I've found another issue @TahaKhanAbdalli 

We should disable the `print receipt` and `print hastener` buttons for default users.  They (mistakenly) try to update the `report printed` dates for the dispatch, which gets rejected by the backend, since this user doesn't have write access to dispatches.

src/resources/dispatch/DispatchList.tsx
```diff
@@ -88,19 +90,19 @@ const BulkActions = (): React.ReactElement => {
       type: 'success'
     })
   }
-
+  const rolesThatCanEditPassword = ['rco-user', 'rco-power-user']
```
IanMayo, line 93: this const shoudl be named `rolesThatCanCreateReceiptNote`

src/resources/dispatch/DispatchShow.tsx
```diff
@@ -147,50 +148,53 @@ const Footer = (props: FooterProps): React.ReactElement => {
   }
 
   if (typeof record === 'undefined') return <></>
-
+  const rolesThatCanEditPassword = ['rco-user', 'rco-power-user']
```
IanMayo, line 151: This `const` needs to be renamed, too.

IanMayo (approved):
Tested and working

IanMayo (approved):

PR #1120, opened 2024-05-22
log item removal from destruction issue resolved

### LINKED ISSUE #1119: Log item removal from destruction
When we remove an item from a destruction, we log the change for the `item`, but not for the `destruction`.

This means that when we view the `history` for a `destruction` it doesn't show the item being removed.

Note: we do handle this correctly for `dispatches`

TahaKhanAbdalli:
Fixes #1119


IanMayo:
I just removed an item from a destruction.  Then I viewed the history for the destruction.  It now included the `item removed` line.  But, when I tried to click on the `item` in this line, the API logged me out.

I see we were still trying to pull data from the `user` endpoint (this is now `_users`).  I guess this is in the `itemShow` page.

IanMayo (approved):
Tested and working

PR #1126, opened 2024-08-02
fix: clear selected items on leaving items page.

### LINKED ISSUE #1125: Clear selection on leaving items page
A user selected some items, then performed an `Action` on them.

The user then went to another page, made some changes, and came back to the `items` page.  They selected an item they needed to perform an `Action` on, and clicked on the `Action`.

But, the previous items were still selected from the previous visit.

So, the users _think_ the correct solution is for the current selection to be cleared when leaving the `Live items` or `All items` pages.

TahaKhanAbdalli:
fixes #1125

### DIFF
--- a/src/resources/items/ItemList.tsx
+++ b/src/resources/items/ItemList.tsx
@@ -639,6 +639,11 @@ export default function ItemList(
       }
     }
   }
+  useEffect(() => {
+    return () => {
+      localStorage.removeItem('RaStore.items-items-list');
+    }; 
+  }, [location.pathname]); 
 
   useEffect(() => {
     if (configData?.projectName) {


IanMayo:
Hello @TahaKhanAbdalli - we need this fix to work on both the `All items` and `Live items` pages.  Do they both use the same cookie?

TahaKhanAbdalli:
Hi @IanMayo, Yes it is working for both `All items` and `Live items` pages.

IanMayo:
Thanks - just one issue. If I switch between `Live items` and `All items` the items remain selected.

The selection should be cleared in this scenario.

TahaKhanAbdalli:
okay let me check.

IanMayo (approved):
Tested and working

PR #1127, opened 2024-08-02
fix: add link icon for items and items details show on new page.

### LINKED ISSUE #1124: Don't clear search filters on page change
When a user goes to an item page from the `Live items` and `All items` pages, then returns to the listing page, the search filters will have been cleared.

Users sometimes have to create complex filters to find the items of interest. So it is frustrating to have to re-enter them.

The search filters are only forgotten on these two pages. They appears to remain in place for other listing types.

So, we should remove the functionality that clears the filter.

A _fallback_ solution would be to convert the `preview` icon into a clickable URL.  Then users can ctrl-click the icon to open the `item` in a new page - thus leaving the search results unchanged.

TahaKhanAbdalli:
fixes #1124

### DIFF
--- a/src/components/DatagridConfigurableWithShow.tsx
+++ b/src/components/DatagridConfigurableWithShow.tsx
@@ -9,7 +9,7 @@ import {
 } from 'react-admin'
 import useRowClick from '../hooks/useRowClick'
 import { type ResourceTypes } from '../constants'
-import { Preview } from '@mui/icons-material'
+import { Preview, Link } from '@mui/icons-material'
 import { ButtonBase } from '@mui/material'
 
 const PreviewButton = ({
@@ -22,17 +22,21 @@ const PreviewButton = ({
   const createPath = useCreatePath()
 
   const handleClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
-    if (event.ctrlKey || event.button === 1) {
-      window.open(`/#/${resource}/${id}/show`, '_blank')
+    if (resource === 'richItem') {
+      if (event.ctrlKey || event.button === 1) {
+        window.open(`/#/${resource}/${id}/show`, '_blank')
+      }
     } else {
-      const path = createPath({ resource, type: 'show', id })
-      redirect(path)
+      if (!(event.ctrlKey || event.button === 1)) {
+        const path = createPath({ resource, type: 'show', id })
+        redirect(path)
+      }
     }
   }
 
   return (
     <ButtonBase onClick={handleClick}>
-      <Preview />
+      {resource === 'richItem'  ? <Link /> : <Preview />}
     </ButtonBase>
   )
 }
--- a/src/components/DatagridConfigurableWithShow.tsx
+++ b/src/components/DatagridConfigurableWithShow.tsx
@@ -23,6 +23,7 @@ const PreviewButton = ({
   const handleClick = (): void => {
     if (resource === 'richItem') {
       window.open(`/#/${resource}/${id}/show`, '_blank');
+     return
     }
       const path = createPath({ resource, type: 'show', id });
       redirect(path);    
--- a/src/components/DatagridConfigurableWithShow.tsx
+++ b/src/components/DatagridConfigurableWithShow.tsx
@@ -20,15 +20,12 @@ const PreviewButton = ({
   const { id } = useRecordContext()
   const redirect = useRedirect()
   const createPath = useCreatePath()
-
-  const handleClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
+  const handleClick = (): void => {
     if (resource === 'richItem') {
       window.open(`/#/${resource}/${id}/show`, '_blank');
     }
-    if (!(event.ctrlKey || event.button === 1)) {
       const path = createPath({ resource, type: 'show', id });
-      redirect(path);
-    }
+      redirect(path);    
   };
 
   return (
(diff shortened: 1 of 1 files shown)


IanMayo:
Hello @TahaKhanAbdalli - this doesn't really work for me.

I see the new icon, but I can't get to ItemShow pages.

TahaKhanAbdalli:
The itemShow page show when the user ctrl+click on this link icon.

IanMayo:
Aah, I see.  But, there is no visual clue that this will happen.   I think we need it to a be a proper `<a>` element, so when the user right-clicks on it, they see "open in new page" (which will effectively also work for ctrl-click).  Obvs, clicking on the `<a>` link will navigate to that page, so we'll have to educate the users to open in a new tab.

But, I _suspect_ this strategy is a less effective than removing the "clear search terms" action.  Other list pages do not lose their search terms when drilling down to a child item.

TahaKhanAbdalli:
Is it Correct? If the user hover in link icon it will show the `Tooltip` with message `Open in new Tab`.

Screenshot from 2024-08-02 18-27-39

IanMayo:
I see your point, but I didn't "linger" long enough to see the tooltip.  The next user may not either.

TahaKhanAbdalli:
Okay let me check.

TahaKhanAbdalli:
Hello @IanMayo,
I have a question regarding the filter functionality on the item list. If a user applies a filter on `ricItems`, should those filters also be applied and visible on `allItems`?

IanMayo:
> Hello @IanMayo, I have a question regarding the filter functionality on the item list. If a user applies a filter on `ricItems`, should those filters also be applied and visible on `allItems`?

Good question.  No, it does not need to be applied and visible on `allItems`.

TahaKhanAbdalli:
Hello @IanMayo,

I have been working on the filter issue and have made some changes. With the current implementation, if a user applies filters on `richItems` or `allItems`, performs actions on the filtered items, and then come back  to the list, the filters will persist. However, if the user switches between different resources, the filters will disappear.

If you agree with this solution, I will push the code. Otherwise, we will need to update all functionalities related to item filters in the audit log list and various other places.

Otherwise, we can apply the same functionality to the preview button. Clicking the preview button can open the item details in a new tab preserving the filters in the same tab if user wants to get back the filters.

IanMayo:
Sure - that strategy sounds fine.

I personally would also value the `preview` button being a real hyperlink, that I can right-click on an select "Open in new tab".  In doing this change, it would be acceptable to lose the "select item when click anywhere in that row".  It would be acceptable for just the checkbox to be used for item selection.

(I kind of think that one of our issues has been separating the click events from the preview icon and the rest of the row).

TahaKhanAbdalli:
Hi @IanMayo,

I have also attempted the `hyperlink` strategy, but when opening an item in a new tab using a hyperlink, it required a refresh of cookies. This happens because opening a link in a new tab is treated as a separate browsing context, which can affect cookie sharing based on the browser and its configuration.

Currently, I am using `window.open` to open the item in a new tab. This method typically allows the new tab/window to share the same session cookies as the original tab, thereby preserving the session context.

Therefore, the best approach is to open the item details in a new tab by clicking the preview icon button.

IanMayo:
In mine, the original tab and the new tab both move to the new URL.

Steps to reproduce:
1. Open Live Items
2. Filter for `765` - 4 items shown
3. Click on `Preview` for the first item, View-Item for #1785 opens in a new tab
4. Close the new tab.  The original tab has moved from `List` to View-Item for #1785

We don't have this issue for other resources.  Note this behaviour:
1. Open `Users` tab
2. Filter for `10` - 9 items shown
3. Click on `Preview` for STAFF-4a
4. Click on `Edit` 
5. Slightly modify the name
6. Save the record
7. Click on `Users`
8. Clear the single selection - the `10` filter is still applied

TahaKhanAbdalli:
Hello @IanMayo,

Could you please let me know is it correct?
1. Go to Live Items.
2. Add the filters: "Created By Me" and set the "Batch Filter" to 10.
3. After applying the filters, you should see 2 items.
4. Open these items in a new tab by clicking on the preview icons.
5. Click the edit button for both items, one by one.
6. Make some changes to the items and save them.
7. Return to the original tab where you applied the filters.
8. Un-select the items by check box.
9. The filters still be there.
10. If i refresh the page the filter will disappears.

TahaKhanAbdalli:
Yes, I haven't push the code yet. I am just confirming is it correct or not?

Go to Live Items.
Add the filters: "Created By Me" and set the "Batch Filter" to 10.
After applying the filters, you should see 2 items.
Open these items in a new tab by clicking on the preview icons.
Click the edit button for both items, one by one.
Make some changes to the items and save them.
Return to the original tab where you applied the filters.
Un-select the items by check box.
The filters still be there.
If i refresh the page the filter will disappears.

IanMayo:
To be honest, you shouldn't have to clear the `is selected` flags.  Clicking `preview` should not mark the item as selected.

I suspect this just requires us to remove the click handler on the data grid:
image

Note: removing the click handler _may_ mean we no longer need to declare a custom datagrid component (`DatagridConfigurableWithShow`).

TahaKhanAbdalli:
Hello @IanMayo. 
Please check this is it correct?

https://github.com/user-attachments/assets/aeefdbab-d3a1-4438-96a3-9ba7f0eadd1b

IanMayo:
Ideally, the two items would not be selected by clicking on `preview`.

But, it's correct that the first tab stays on the `List` view.  I don't know why mine doesn't.

TahaKhanAbdalli:
I have made some changes to the code, which is why it is now working correctly on my system. Let me push the code then please verify me is it correct or not.

IanMayo:
Hello @TahaKhanAbdalli - I've removed the `rowClick` behaviour.  Now, items are only selected by clicking on the tick-box. This is the same behaviour as the other `List` pages.

IanMayo (approved):
Working correctly

PR #1128, opened 2024-08-02
add extra guidance

### LINKED ISSUE #1122: For add-user, need to indicate password complexity tests
A power-user was adding a new user, but the new (temporary) password wouldn't be accepted.

She had forgotten about the password complexity rules.

We should include the password complexity rules near the form control for the temporary password.

IanMayo:
Fixes #1122

### DIFF
--- a/_devExtensions/password-validation.schema.js
+++ b/_devExtensions/password-validation.schema.js
@@ -94,7 +94,7 @@ const passwordValidationSchema = yup
   .matches(/[0-9]/, 'Password must contain at least one numeric digit')
   .matches(
     /[!@#$%^&*(),.?":{}|<>]/,
-    `${PASSWORD_SPECIAL_CHARACTER}`
+    'Special character required - !@#$%^&*(),.?":{}|<>'
   )
   .test(
     'no-common-patterns',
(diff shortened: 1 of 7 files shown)


TahaKhanAbdalli:
I believe we should display an error message if the user does not enter a password correctly. Currently, users are being created even if the password length is only 8 characters.

IanMayo:
> I believe we should display an error message if the user does not enter a password correctly. Currently, users are being created even if the password length is only 8 characters.

Great - would you mind adding that please?

IanMayo:
Thanks Taha, that works great - I hadn't realised we weren't doing pre-checking on the page.

I remember we had feedback that requested that we include the acceptable `special characters` in the warning message

I'd like you to add that, please.  But, on looking at the code, I see we do provide them in the warning message in one place:
image

I also see that we have multiple instances of the password checking code, and they are not all the same. I'd like the client-side code to all use just one implementation of password checking logic, please.

I don't mind having separate client and server instances of the password checking code - if necessary.

IanMayo:
@TahaKhanAbdalli - do you know why we use custom password validation in `UserShow.validatePassword`?

IanMayo:
Thanks Taha - can we also declare the explanatory block as a const in just one place too, please?
image

Aah, we could do that in the same place that we declare the validation schema.  Plus, if we do that, we can declare the list of special characters as a constant, and use that constant in both the schema and the explanatory notes :-)

IanMayo:
Thanks Taha, but I get this runtime error.  I presume we need to import the constant.
image

IanMayo:
Great, thanks @TahaKhanAbdalli .

